PR #1, opened 2011-03-08
Added :dependent => :destroy to model relationships

strass:
Added them so that if either a game or a topic is deleted, associated topics and posts are also deleted.


raid5:
Merged in :dependent => :destroy commit. (SHA: 5087cb29395f686afd9f565e1a4d3e50526540b3)

PR #8, opened 2011-05-20
Add locked topics

knewter:
Add a locked? boolean to topics.
Add 'can_be_replied_to?' method to topic that checks this boolean.
Do not show reply links if a topic is locked.
Do not allow posts#create to succeed if its topic is locked.
Add spec to verify posts can't be created on locked topics.


knewter:
Also added classes - 'locked' and 'unlocked'

PR #11, opened 2011-05-20
[add views] Add view model, register a view each topic show, and show vie

knewter:
[add views] Add view model, register a view each topic show, and show view count on topics listing.


raid5:
Wouldn't this make more sense to track the number of views in a single `View` entry per user with a `count` column?

knewter:
Maybe.  I'd also want to keep track of when the views happened, because in a forum there's a lot of value in 'highlight posts since my last visit' etc.

We've sort of tended to prefer inserts over updates based solely on experience.  I'm not terribly concerned about the size of this table (it's a tiny bit of data).  I do need to add in timestamps I realized just now, can't believe I left those out.  Pushed that commit.

Anyway, this gives more data.  I'd be fine with a refactor, but I always prefer more data over less, and massive numbers of simultaneous updates to a table in a very large project we work on led to random and hard to track down row/table locks in the database.  Inserts never cause this.

Anyway, thus went my reasoning at least.

raid5:
Seems like good reasoning to me if this has been your experience :) It is very little data, I wasn't really too worried about that...was just throwing out another idea.

knewter:
No prob at all.  You can also then do nice little rollup graphs (which forums, topics get the most traffic between 2AM-4AM on the weekends?)  But yeah, glad to have someone commenting on my code.  Discussion is fun.  Do you have any interest in hopping in FreeNode#forem ever?

knewter:
So I need to mess with this some more, will re-open the request.  Some specs are failing - don't believe they were before, but it might be an intermittent sort of thing.  Researching.

knewter:
Fixed the tests (bug was introduced by an innocuous-seeming change that I did not test afterwards, which should teach me a lesson but likely won't).  Also added a test to verify that it actually creates a view record.

radar:
This seems like to me that every time a user views a topic that it would insert a record into the database, leaving open the possibility for abuse for a user if they were to discover that's what we're doing. I would much rather it stored just _one_ row for a user and a topic and the created_at timestamp served as kind of a "last viewed at" timestamp. This would be the best of all worlds, I think, as we would have a clear representation of how many _unique_ (user) views we've received for a given topic and we would be able to work out very easily if a topic has received more posts since the user last visited it with a very simple query.

What do you think?

p.s. View tracking is a _hard_ problem to solve.

paxer:
none asked me, but I would agree with Ryan, it does not make sense to add this level of complexity for this task. So - "just one row" is better solution IMHO.

parndt:
I just pushed a commit which ignores the logs from the dummy project, it was affecting me.

parndt:
@knewter status?

knewter:
I guess I need to refactor it to do the view counting in a single row per user....but it sort of doesn't make full sense to me yet.  For my use case, we also want to track views by not-logged-in users - would I just have a single row then without a user id and update that like I would a user's view row?  Seems the best I could do - don't know that it's a terrible idea.  Thoughts?

radar:
I would just have this code in an `update_view` method for `TopicsController`, called from `show`:

```
if current_user
   view = View.find_or_initialize_by_user_id_and_topic_id(current_user.id, topic.id)
   view.update_attribute(:updated_at => Time.now)
end
```

The `update_attribute` method is being used rather than `touch` because I think `touch` tries to do too much, AND `update_attribute` skips validations, where `touch` doesn't.

This will then create / update the view record for this user on this topic and allow us to display an icon if they haven't viewed the latest post on a topic.

As for anonymous users: don't bother.

knewter:
Meh, there's some non-view stuff in here but I wish you'd just merge it :)  This stuff works now as described.  I need some help on the spec because sqlite's being a dick with nested transactions, so there's one pending spec (that works but I can't show it for some reason) and I turned off transactional fixtures, so that might upset you :)  PLZ help with the tests?

PR #19, opened 2011-06-26
more features that missed feature branches

knewter:
Admin: Edit topics. - pin / hide / lock
Topics: Reply with quoted text
Forum index: show total number of views in forum
Forum show: show whether a topic has had new posts since your last visit, if you've visited it.


parndt:
"This pull request can not be automatically merged."

Issue #15, opened 2011-06-07
Use Gravatar for Avatars


radar:
If the User model has an email method, we should use it to display gravatars of those people on the posts.

parndt:
@radar Should we store the avatar or just hotlink gravatar each time?

radar:
I believe GitHub hotlinks them and so if it's good enough for GitHub it's good enough for us.

parndt:
I have this setup on the http://refinerycms.com/blog already so I can whip the code in. :)

radar:
Ship it squirrel says:

![Ship it!](http://i.imgur.com/s7MRg.jpg)

Ship it!

PR #23, opened 2011-07-08
Display gravatar if user has an email address.

jamescook:
For issue #15

p.s. how do you attach code to an existing issue?


radar:
I don't know how you would attach code to an issue, but thanks :)

PR #26, opened 2011-07-23
Fix user class on posts and views

knewter:
Oops, duplicating the hidden migration fix.  Anyway, the associations to user weren't properly genericized, fixed.


radar:
How does this teach it to be more liberal? Sorry, I am just not sure what this change is doing.

PR #27, opened 2011-07-27
Fixed some admin typos in the topics section

STRd6:
Locked probably needs a default => false in the database migration so we won't need to use the cheese !! to coerce it.

Pinned was just typoed as locked.


parndt:
Could you add / change a migration to add the default instead please? Thanks!

parndt:
Can you edit the patch (on github select the patch-1 branch and edit the file) to remove the !! then I'll merge this too?

parndt:
Thank you! :)

PR #30, opened 2011-08-04
Switched first and last on topic listing posts to get the correct one.

STRd6:
I'm not sure if this is the best place for this fix, but I noticed that the latest and started by were reversed.

### DIFF
--- a/app/views/forem/topics/_topic_listing.html.erb
+++ b/app/views/forem/topics/_topic_listing.html.erb
@@ -14,10 +14,10 @@
   </td>
   <td class='topic'>
     <div class='subject'><%= link_to topic_listing.subject, forum_topic_path(@forum, topic_listing) %></div>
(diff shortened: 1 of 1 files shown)


parndt:
Is there a default scope that sets order by created at DESC? I ask because I find it strange that it'd be like this otherwise :)

STRd6:
Not sure, looking at the posts association in the Topic model it seemed like it should be the other way but they end up swapped in the view. Take a look at http://forem.heroku.com/, it exhibits the error.

parndt:
@radar ?

radar:
forem.heroku.com could be exhibiting this error as it hasn't been deployed to in a while. I believe we have tests that are now covering this issue and since all the tests are passing I don't believe that this is an issue. If you can create a test case that reproduces this issue then I'd be happy to review again.

PR #33, opened 2011-08-28
Namespacing migrations

filiptepper:
Hi guys,

Thanks for all your great work!

I ran into an issue when trying to hook up forem with my app, which already has CreatePosts migration. What do you think about namespacing migrations like in my commits?

Cheers!


parndt:
I believe this is necessary and now you have me thinking about other project like refinerycms!

Thanks!

radar:
Thanks guys! :)

filiptepper:
Awesome, thanks Philip!

PR #35, opened 2011-08-30
Polish locale

filiptepper:
Polish locale for forem, coming right up!

### DIFF
--- a/config/locales/pl.yml
+++ b/config/locales/pl.yml
@@ -1,0 +1,99 @@
+en:
+  forem:
+    admin:
+      forum:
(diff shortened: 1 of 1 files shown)


config/locales/pl.yml
```diff
@@ -0,0 +1,99 @@
+en:
```
parndt: this should be `pl` not `en` :D

filiptepper: Geez… ;-)

filiptepper: Fixed!

parndt:
Thanks @filiptepper !

PR #43, opened 2011-09-10
Edited forem.gemspec via GitHub

bearded:
Hi guys,

Please pull

```
`git ls-files`.split('\n').size => 1
`git ls-files`.split("\n").size => 366
```


parndt:
Thanks for fixing my fail :+1:

PR #46, opened 2011-09-13
Just added a german localization yml

cobalamin:
Not much work, but I wanted to contribute a bit :)
Thanks for your work, I based a whole small forum system on mainly forem's logic for my app :)


config/locales/de.yml
```diff
@@ -0,0 +1,98 @@
+en:
```
parndt: This should be de

Issue #49, opened 2011-09-22
Getting ActiView Template Error


gabrielbaez:
Hey,

I followed your instructions and installed forem,  but when I try to view the forum  I get the following error message.

ActionView::Template::Error (undefined local variable or method `blogs_path' for #<#Class:0x7f4f21002090:0x7f4f20ffda40>):

I created the initializer per the instruction  and  configured the application controller 

App controller code  http://pastie.org/private/nuf4t0cz11dscgirz4kpq 

This is what I currently  have on my gem file

```
source 'http://rubygems.org'

gem 'rails', '3.1.0'

# Bundle edge Rails instead:
# gem 'rails',     :git => 'git://github.com/rails/rails.git'

gem 'pg'

gem 'devise', '1.4.5'

gem "cancan"
gem 'json'

gem 'ckeditor', :git => 'git://github.com/galetahub/rails-ckeditor.git', :branch => 'rails3'

gem 'forem', :git => "git://github.com/radar/forem.git"
gem 'forem-theme-base', :git => "git://github.com/radar/forem-theme-base.git"

gem 'kaminari'

gem 'gravatar_image_tag'

gem 'activeadmin'
gem 'sass-rails', "  ~> 3.1.0"
gem "meta_search",    '>= 1.1.0.pre'
gem 'therubyracer'

gem 'paperclip'


# Gems used only for assets and not required
# in production environments by default.
group :assets do
  #gem 'sass-rails', "  ~> 3.1.0"
  gem 'coffee-rails', "~> 3.1.0"
  gem 'uglifier'
end

gem 'jquery-rails'

# Use unicorn as the web server
# gem 'unicorn'

# Deploy with Capistrano
# gem 'capistrano'

# To use debugger
# gem 'ruby-debug'
```

radar:
This is because in the layout for your application you're using `blogs_path`, when you should be using `main_app.blogs_path`. The engine will attempt to look up `blogs_path` in its own routes (because its isolated), not find it and bomb out with that error. That is even though your application has it defined!

You _must_ do this now if you are using engines, as this will enable your application and engine to remain isolated (which allows them to have identically named routes and not conflict) but still have some kind of inter-connectivity between them. To access forem routes within your application you must use `forem.root_path` and the like.

Thanks for using Forem!

gabrielbaez:
Thanks man didn't know that,  that fixed it!

PR #51, opened 2011-10-01
adds engine route helper documentation from issue #49

nruth:
I lost about an hour to this, then found it in the issues tickets. 
It'd be cool to have it in the readme, so I wrote something brief.

What do you think?

Cheers,
Nick


radar:
Perfect, thank you :)

PR #52, opened 2011-10-01
removes and ignores .rvmrc

nruth:
which should be developer preference not project setting, as common in other repos.

I think this was probably added to git by accident?

The only quirk with this commit is it will probably delete your local .rvmrc file, but it's easy to recreate it?


radar:
That's fine if it deletes it. Thanks for the patch, you're right :)

Issue #55, opened 2011-10-05
Textile / pluggable post formatting


nruth:
We'd like to use textile rather than markdown for post formatting (for consistency with our cms).

Does anyone have suggestions for how this could be architected / an appropriate pattern for having plain text, or markdown, as default, but the possibility to change to something else?

radar:
This is a very similar problem to #50, where the users would like to use WYSIWYG formatting. I suggested there that they split the part of the post partial that renders the text using the `forem_markdown` helper out into its own partial, which could then be over-written in the application (by simply having a partial called the same) which would allow you to format it using Textile.

But I think this solution may be a bit hacky. Do you have any other ideas on how you would do this?

nruth:
That's a good point. We've already used the generate views stuff so I'll look at this as a quick-fix.

I was thinking of the adapter pattern, having a (forum) post processing adapter, which we could plug in and out as wanted.
The interface seems pretty simple, one method which takes the post text and returns HTML.

This would probably allow someone to extend markup to be a user preference, if the controller and user model were made smart enough. This isn't something I need right now but it could be attractive.

A residual problem would be the default gem dependency for markdown. This doesn't bother me much, but it does give me another idea. If no formatting was the default, then markdown, textile, or whatever, could be installed by adding a gem, like for loading themes.

radar:
The adapter idea is a pretty good idea. 

The `forem_markdown` helper should be renamed to `forem_format`, which would simply call `Forem::Post#format`. This method then should be defined as just returning the post text, i.e. a no-op.

The markdown / textile / wysiwyg functionality then would be provided by a gem that has a module which is included into `Forem::Post` which would override this method, causing the text to be formatted correctly. As for the form, you could probably also override that in the engine, but I am not confident that would work as well as having it in the application.

Please do investigate :)

nruth:
I've made the initial core changes, and put together a [gem for RedCloth/Textile formatting](https://github.com/nruth/forem-textile_formatter). It's also possible to just override the helper in the app which is loading forem.

https://github.com/nruth/forem/commit/486c1e54e834c8f2622ccdba255a9c4918aa5381

I'm not sure returning plain text is a good idea though; people will expect (safe) html to come back, which they can just insert into their template. Do you think it's sensible to make the default behaviour to wrap plain text in a pre tag? 
Or a p tag? More than that gets into re-implementing markdown territory.

nruth:
[...]

PR #62, opened 2011-10-14
Removes Markdown default, replaces forem_markdown with forem_format

nruth:
From issue #55
- Adds basic tests for default formatter.
- Removes redcarpet and markdown refs.
- Deprecates forem_markdown helper, no longer produces markdown and logs warning
- Refers to gem plugins which give Textile or Markdown formatting

Default formatter:
- wraps text with pre tag
- escapes unsafe post html

To replace formatter with markdown / textile override Forem::FormattingHelper#as_formatted_html


knewter:
Nice code.  Merges cleanly into master atm.  Works as advertised.  Is test-covered.

PR #63, opened 2011-10-14
test for "can't delete other people's posts"

nruth:
I noticed this was deleted in a previous commit, likely out of frustration @ capybara not doing what was expected with the delete request (it drops the rack-test session, losing the log-in, or something like that).

This version of the test passes.


knewter:
This code looks fine and the tests pass just dandy

radar:
Thanks for submitting a fix for this!

PR #64, opened 2011-10-18
Portuguese translation

caffo:
Hi, I've just translated forem to brazilian portuguese. I would love to see it merged to the main repo.

Thanks!


config/locales/pt-BR.yml
```diff
@@ -0,0 +1,116 @@
+en:
```
parndt: This should be pt-BR

caffo:
Fixed.

Issue #60, opened 2011-10-12
Translations of ActiveRecord models and attributes


LegendSzymon-P:
Models names and attributesm such as Topic and Post, are not translated - for example in a new topic form, i have "Subject" and "Text" instead of translated into in my case Polish.

But even when I add such as translation to my config/locale/pl.yml:

```
pl:
  activerecord:
    attributes:
      topic:
        subject: Temat
        text: Tekst
```

I still have `Subject` and `Text` as labels in new topic form. Why?

radar:
Does `simple_form` attempt to reference translations by default automatically? If not, then that is probably the problem. Could you submit a pull request that makes these fields use the translations?

Thanks!

LegendSzymon-P:
Ok, i've find out how simple_form is looking for attributes names and I've added polish translations to them in this pull request: https://github.com/radar/forem/pull/65

But there is still one problem - on the new and edit forms of topic there is a text on a button which uses name of model (Create Topic and Update Topic). I've find out so far that this is not forem thing, but simple_form, and it is using default rails-i18n paths for it: `helpers.submit.create` and `helpers.submit.update`.

I've finally just added translations of these buttons for each model: `helpers.submit.model.action`, but still I don't know how to translate models names which are used in these buttons, even when I add `activerecord.models.model` translation, it still is loading english name of model to these buttons, I don't know why, maybe because forem models are in `Forem` module?

parndt:
@apocalyptiq Was this fixed by #65?

LegendSzymon-P:
Yes, all is alright now.

PR #65, opened 2011-10-20
Polish translations of models attributes and a few other and indicators to them in english translation

LegendSzymon-P:
I've added polish translations of models attributes and indicators to them in english translation, so that when somebody will update some other language translation, he will new what to translate :)

I've mentioned that problem previously in that issue: https://github.com/radar/forem/issues/60

I've find out in simple_form readme that it looks for attributes like this: `simple_form.labels.model.attribute`

But there is still problem with models names translations, but I will describe it on that previous issue

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -20,6 +20,9 @@ en:
         topics: Topics
         posts: Posts
         delete_confirm: Are you sure you want to delete this forum?
+        index:
+          last_post: Last post
+          none: None
     common:
       pages: Pages
     errors:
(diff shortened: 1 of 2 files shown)


LegendSzymon-P:
Ok, i've found out how to solve that problem with models names - just overwrite whole label of button for each model :) like this: `helpers.submit.model.[create|update]` , but I still don't know how to achive that only translating models names, but I think that solution is ok.

parndt:
Ready to merge?

LegendSzymon-P:
@parndt yes

PR #70, opened 2011-10-26
Indexes

idearise:
I'd like to add indexes to the forem tables. It also has the minor topics_helper change. Thanks!


parndt:
Which version of bundler are you using?

idearise:
bundler (1.0.21)

PR #76, opened 2011-11-11
Added Ability to move  topic to a different forum. 

gabrielbaez:


app/views/forem/admin/topics/_form.html.erb
```diff
@@ -1,5 +1,6 @@
 <%= simple_form_for [:admin, @topic] do |f| %>
   <%= f.input :subject %>
+  <%= f.association :forum , :label => "Forum" %>
```
gabrielbaez, line 3: I had to specify the label here  because without it, it was outputting "Title,  description"  text.

radar:
Could you add a test for this please?

gabrielbaez:
I can, but it is a bit difficult since rake routes  does not  work  to get the list of  paths  
am I running this wrong ?  
gabriel@railsdev:~/forem$ rake routes
rake aborted!
Don't know how to build task 'routes'

(See full trace by running task with --trace)

reset:
@gabrielbaez when working with an engine that uses a dummy for testing you have to access the Rake tasks under the app namespace

`rake app:routes`

Still, you might not get the output you are expecting since Forem is a fully isolated engine.

radar:
That will give you the dummy application's routes, but not the engines. I've just added `rake forem:routes` in @dabb3b62d7b4f648d8799499992b605f7e5cf6d5 that will give you forem's routes.

reset:
Weird that theres no built in way to get an output of an Engines routes in Rails yet

gabrielbaez:
Thanks for the help,  added the test let me know how it looks it passed on my end.

gabrielbaez:
@radar  added the check,  thanks.

radar:
Great, thanks for your work!

PR #79, opened 2011-11-17
Add support for blockquotes coming from formatters

jgadbois:
This will require formatters to implement a blockquote method.


app/helpers/forem/formatting_helper.rb
```diff
@@ -8,5 +8,13 @@ module Forem
         simple_format(h(text))
       end
     end
+
+    def as_quoted_text(text)
+      if Forem.formatter
```
radar: How about this instead?

```
if Forem.formatter && Forem.formatter.respond_to?(:blockquote)
```

That would mean that you wouldn't break existing formatters and future formatters wouldn't need to implement a blockquote method.

jgadbois: Yeah, I like that - committed

parndt:
Can we please have tests for this functionality?

Thanks!

jgadbois:
Started on some specs which brought up a problem - all the formatters run the raw markdown through ERB::Util.h which escapes the markdown blockquote formatting (start the line with greater than).

The current commit in the pull request gets around this by wrapping the post in raw() but that's probably not smart and causes an existing spec to fail.

I'm pretty sure the blockquote tag gets escaped using the current functionality too.

Any suggestions?  I was thinking there might be some markdown sanitizer we could use instead of h()?

parndt:
What do you think @radar ?

radar:
Ok, perhaps we should be running the formatter and then running that output through a HTML sanitizer. Any experience with this @parndt?

parndt:
@radar none

jgadbois:
I removed the escaping from the RDiscount gem and added a call to sanitize before outputting the forem post.  Specs (my new ones and existing ones) are passing.

jgadbois:
Sorry some of these pull request commits are on my master in addition to my formatted-quotes branch, so I'll have to have this pull request merged before I can fork off another branch to start working on the email subscriptions.  I'm pretty new to pull requests so sorry for messing that up.

jgadbois:
It should - I'm sanitizing before running through markdown.

radar:
Alright, I'll run through it tonight when I'm home from work.

jgadbois:
@radar - sorry to bug you.  Had a chance to check this out yet?

radar:
No sorry I've not had the time, yesterday's work took longer than I thought and I also spent the morning + evening biking 50km.  

I am in Adelaide this weekend (until the 27th) visiting my family. I will try to find some time to look into it this weekend.

radar:
There's a binary file in this called "core". What is that?

radar:
This seems to not work... when I click "Quote" on a post it gives me the `<blockquote>` wrapping for the text, but when the post is created this tag vanishes.

radar:
I've made [a new commit to forem-rdiscount here](https://github.com/radar/forem-rdiscount/commit/84470d4) which seems to fix this error.

jgadbois:
Thanks for committing this - I'll get started on email notices.

PR #81, opened 2011-11-23
Gravatar configuration

phiggins:
Allow setting a default theme for gravatars or a static default image.


radar:
Looks clean, thanks for the work on this!

PR #82, opened 2011-11-24
User profile links

browning:
Gives the option to make the username text next to a post be a link to some user profile page that the developer chooses.

### DIFF
--- a/app/views/forem/posts/_post.html.erb
+++ b/app/views/forem/posts/_post.html.erb
@@ -2,7 +2,13 @@
 <div id='post_<%= post.id %>' class='post <%= cycle('odd', 'even') -%>'>
 
   <div class='user'>
-    <div class='username'><%= post.user -%></div>
+    <div class='username'>
+      <% if not Forem.enable_user_profile_links %>
+        <%= post.user -%>
+      <% else %>
+        <%= link_to post.user, [main_app, post.user] %>
+      <% end %>
+    </div>
     <div class='icon'><%= avatar(post.user, :size => 60) %></div>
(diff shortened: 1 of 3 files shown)


app/views/forem/posts/_post.html.erb
```diff
@@ -2,7 +2,13 @@
 <div id='post_<%= post.id %>' class='post <%= cycle('odd', 'even') -%>'>
 
   <div class='user'>
-    <div class='username'><%= post.user -%></div>
+    <div class='username'>
+      <% if not Forem.enable_user_profile_links %>
```
parndt, line 6: In Ruby, we have the unless keyword instead of "if not"

And actually for this case you'd be better off with:

link_to_if Forem.enable_user_profile_links, post.user, [main_app, post.user]

lib/forem.rb
```diff
@@ -2,7 +2,8 @@ require 'forem/engine'
 require 'kaminari'
 
 module Forem
-  mattr_accessor :user_class, :theme, :formatter, :default_gravatar, :default_gravatar_image
+  mattr_accessor :user_class, :theme, :formatter, :default_gravatar, :default_gravatar_image, 
+    :enable_user_profile_links
```
parndt, line 6: Could the setting be called user_profile_links (aka without "enable")? I find it reads better.

browning:
Thanks a lot for the advice, going to close this pull request and do a new one with these changes

parndt:
Or you can just push more commits to your fork to the same branch
which automatically updates the pull request

PR #83, opened 2011-11-24
Adding user profile link option to posts

browning:
Same idea as last time, but using some good suggestions from parndt


parndt:
@radar should we get a test for this too?

radar:
it's alright, I'll do the test myself. Thanks for the pull request browning!

Issue #53, opened 2011-10-02
Authorization system


radar:
Last week, I spoke with somebody was interested to know if Forem had any kind of authorization system. I told him it didn't, because I didn't know how to build one for it as applications can have their own authorization system.

So I spent some time thinking about this on the plane ride back from Chicago. Here are my thoughts in no particular order. It would be great if I could get some feedback on this issue:
### Permission system for forem

In many other forum systems out there, there are permission systems built into them. Forem is unique in that it's an engine and supposed to be pluggable into other applications, rather than being something standalone such as (my understanding) how phpBB and vBulletin work.

Of course, Forem can be "standalone" by just mounting it inside an empty Rails application. I don't think there will ever be  a case where we have a true standalone version.

Anyway...

This complicates things. By creating our own permission system we would limit those people who may have their own permissions systems within their applications. This is definitely not ideal. Forem is already able to adapt to different _authentication_ systems, so why not _authorization_ also?

I propose that we build a flexible authorization system into Forem so that people can set up permissions on their forem installs in the easiest way possible using their own authorization if that's what they have, or suggesting to use CanCan instead.
#### How it works

I think that we should have methods on the `User` class which define the permissions that people have. For example, there would be a `can_read_forum?` method which would take one argument and do something like this:

```
def can_read_forum?(forum)
  permissions.where(:object => forum, :permission_type => "read")
end
```

The code inside the `can_read_forum?` method will be application-specific, but the API for the permissions is something that we should provide.

There could even be a possibility that we provide a `Forem::Permissions::Default` module which, when included into a `User`-like class, would then provide defaults for these permission methods, which then can be overridden in the model as the application designers wish. I think it would be best to have a blacklist-by-default, but I am open to ideas on this one on why it's a bad idea if anybody thinks so.
#### Alternative API

Alternatively we get them to include a module called `Forem::Permissions` which then provides the following API:

   can_read_forum do
     permissions.where(:object => forum, :permission_type => "read")
   end

This then will store the given block which will be executed (using `instance_eval`) on the `current_user` object.
#### Caveats

With this kind of permission system I cannot think of a clean way that you could allow anonymous access to forums (for example). How could we fix that?

Thanks lads!

nruth:
Sounds cool.

I think it'd be convenient to have blacklist and whitelist defaults available for default permissions.

For anonymous access, what about a non-persisted object to represent anonymous visitors? 
It could have permissions set in the class def.

tanraya:
[...]

PR #92, opened 2011-12-05
Let users edit their posts

browning:
Hey, just a change to let users edit their own posts. Also lets the admin edit any post.
Let me know if you think we should do it differently.
Thanks!

### DIFF
--- a/app/controllers/forem/posts_controller.rb
+++ b/app/controllers/forem/posts_controller.rb
@@ -31,11 +31,13 @@ module Forem
     end
 
     def edit
+      authorize! :edit, :all
       @topic = Topic.find(params[:topic_id])
       @post = Post.find(params[:id])
     end
 
     def update
+      authorize! :edit, :all
       @post = Post.find(params[:id])
       if @post.owner_or_admin?(forem_user) and @post.update_attributes(params[:post])
         redirect_to [@topic.forum, @topic], :notice => t('edited', :scope => 'forem.post')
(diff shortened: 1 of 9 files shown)


app/controllers/forem/posts_controller.rb
```diff
@@ -30,6 +30,22 @@ def create
...
+    def edit
+      @topic = Topic.find(params[:topic_id])
+      @post = Post.find(params[:id])
+    end
+
+    def update
+      @post = Post.find(params[:id])
+      if @post.owner_or_admin?(forem_user) and @post.update_attributes(params[:post])
+        flash[:notice] = t("forem.post.edited")
+        redirect_to [@topic.forum, @topic]
```
parndt: Can we use this here instead?

``` ruby
redirect_to [@topic.forum, @topic], :notice => t('edited', :scope => 'forem.post')
```

That sorta redirect code is built into Rails now

radar: I agree, probably better.

radar:
Hey, thanks for the terrific patch!

There's a comment [https://github.com/radar/forem/issues/53](here) that asks if we could have permissions determining if users could edit posts. Do you reckon you could write a little more into this patch to support that?

That'd be awesome! :)

browning:
Yep, I will take a look at adding permissions tonight

browning:
Alright, I think this is good to go now

radar:
Neat work! Hooray for awesome contributors!

I just made a small change in @01c3a39 to allow people to customise where users can edit posts, so that some users can edit posts in some forums but not others if they feel like it.

PR #96, opened 2011-12-12
remove 'text' label on the new post form, its not really needed

browning:
Hey guys, right now on the create post form there is a label that just says 'Text', it looks kind of out of place and I don't think we need it so this just removes it

### DIFF
--- a/app/views/forem/posts/_form.html.erb
+++ b/app/views/forem/posts/_form.html.erb
@@ -1,4 +1,4 @@
-<%= f.input :text %>
+<%= f.input :text, :label => false %>
 <% if params[:reply_to_id] %>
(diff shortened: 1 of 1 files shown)


browning:
actually sorry, hold on, removing this label breaks some of the tests
Looking into fixing the tests now

radar:
The label is used in the tests to know what field to fill out. To fix these tests you would need to change "Text" to "post_text" in them.

I don't think removing the label is a good idea, but I could probably be convinced.

browning:
Yea I will close for now. 
With it removed the new topic page looks worse, but the reply page looks better IMO.
Will try to come up with a better solution if I have some free time later.

PR #99, opened 2011-12-16
Fix error on forum index page

gabrielbaez:
I had to remove the "only" option from the load authorize resource  it was preventing me from viewing the forum
index.

It was triggering the following error.

CanCan::AuthorizationNotPerformed in Forem::ForumsController#index

This action failed the check_authorization because it does not authorize_resource. Add skip_authorization_check to bypass this check.


radar:
There's quite a large number of commits on this pull request. Shouldn't there only be one? What happened?

Please re-submit as a pull request that only has a few commits.

gabrielbaez:
They are not mine, all I did was  a pull to my  fork to sync up the latest forem changes.  If you look at the diff I  only changed one file.

radar:
Could you please provide some information as to how to duplicate this error? I am unable to do it on my local install.

gabrielbaez:
I ran into this error since my website makes use of cancan.

gabrielbaez:
The same error is also triggered on the admin section,  but on the rest of the forum it seems to be working fine so far.

radar:
Right... but I still don't know how exactly I would reproduce this error... How would I go about doing that if I had a blank Rails application?

gabrielbaez:
When you setup a new rails up that uses cancan and you enable  check authorization in applications controller.  it reproduces the error when viewing the forem index, if you go directly to the show action the error is not triggered. 

 check_authorization :unless => :devise_controller?
  rescue_from CanCan::AccessDenied do |exception|
    flash[:alert] = exception.message
    redirect_to root_url
  end

Issue #101, opened 2011-12-21
Email Subscriptions Proposed Feature List


jgadbois:
Current feature plans (italics is @radar reply to my initial email)
### Subscription Management
- Subscribe/Unsubscribe link on topic view if logged in
  
  _I think this is fine. It should indicate clearly if a person is subscribed or not._
- Either a check box or automatic subscription when adding a post to a topic
  
  _Determining if a user should be automatically subscribed could be done with a forem_auto_subscribe field on the "users" table that defaults to true. This could then be toggled on a profile page somewhere for the user, perhaps? We don't have somewhere like that in Forem at the moment._
- Initially, an unsubscribe link in the email footer (vs a management page for subscriptions?)
  
  _That would be fine, I think._
### Subscription Emails
- Emails get sent out every reply to everyone but the person who added the new reply
  
  _That's correct. My concern with this is that it won't work at scale, but that's something to just think about for the future. There should of course be only one email per person._
- Text of post is in email?

_Not sure how you'd go about preserving the formatting here. I think just a notification of "Blah has posted in your topic called 'blah'" is fine, for now. If you want to go ahead and put the post text in the email then be my guest. I'd actually prefer it that way._

jgadbois:
I implemented the subscribe/unsubscribe link with specs - should I open a pull request even though I am not done so you can comment on the code as I work on it?  I used a Subscriptions controller for subscribing/unsubscribing.  Actions on the topics controller might be cleaner although less RESTful.

knewter:
@jgadbois what's the situation with this one now.  I saw you closed the old pull request, but did you ever open a new one?

radar:
Finished afaik.

PR #108, opened 2011-12-28
Email subscriptions

jgadbois:
Work on issue #101


radar:
Why are there SO MANY commits on this pull request? What happened, mate?

radar:
Actually I think I may know what happened. You forked from master before I did the force push. If you pull origin/master to a new branch called email_subscriptions for example, and then cherry-pick your commits across onto this new branch and then submit a new pull request it should only have the right commits.

jgadbois:
Ahh...I was confused about that too.

I'll try that.

jgadbois:
Resubmitted - that should be a lot cleaner.  Still a work in progress but looking for early feedback.

PR #109, opened 2012-01-01
First Farsi( Persian) Locale for Forem

Amirhb:
This is the first Farsi( Persian) translation for forem( fa.yml). Farsi is an rtl language. I had to add rtl to my local copy of layout to make the Persian( Farsi) translation neat.

### DIFF
--- a/config/locales/fa.yml
+++ b/config/locales/fa.yml
@@ -1,4 +1,4 @@
-en:
+fa:
   forem:
     access_denied: "شما به انجام این عملیات مجاز نیستید."
     admin:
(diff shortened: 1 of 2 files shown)


config/locales/fa.yml
```diff
@@ -0,0 +1,163 @@
+en:
```
radar: This should not be en, but rather fa.

radar:
Could you please also update the README to contain mention of Persian in the languages section? Thanks.

radar:
Merged, thank you!

PR #112, opened 2012-01-02
Adds classes to table elements for advanced theming.

robyurkowski:
This is intended to provide enhanced theming capability, especially for mobile devices, by allowing stylists to hide unimportant columns.


radar:
:heart: :beer: :+1: :rainbow:

Issue #91, opened 2011-12-02
Installer should insert a forum


radar:
The installer, once complete, should create a test forum with a topic created by a dummy user. Otherwise the forum looks empty.

knewter:
@radar think this one's closed now as well

PR #120, opened 2012-01-07
Installer creates a forum and topic

browning:
Hey guys,
I took a stab at handling issue #91, making the installer create a default forum & topic.
Let me know if you have any suggestions.

The one thing thats kinda weird about this is that I am just using the first user in the db as the user that creates the topic & forum, I think its better than creating a new user in the db specifically for this, but I'm open to other approaches.

thanks

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,0 +1,10 @@
+user = Forem.user_class.first
+unless user.nil?
+  forum = Forem::Forum.find_or_create_by_title( :category_id => Forem::Category.first.id, 
+                               :title => "Default",
+                               :description => "Default forem created by install")
+  topic = Forem::Topic.find_or_create_by_subject( :forum_id => forum.id,
+                               :user_id => user.id, 
+                               :subject => "Welcome to forem!", 
+                               :posts_attributes => [{:text => "Hello World", :user_id => user.id}])
+end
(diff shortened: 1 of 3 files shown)


db/seeds.rb
```diff
@@ -0,0 +1,10 @@
+user = Forem.user_class.first
```
radar, line 1: Forem.user_class should be a string, rather than a class. You will need to call constantize here.

db/seeds.rb
```diff
@@ -0,0 +1,10 @@
...
+unless user.nil?
+  forum = Forem::Forum.find_or_create_by_title( :category_id => Forem::Category.first.id,
```
radar, line 3: Is there always a category created during the install?

browning, line 3: Yes , as long as migrations are run, the general category gets created in the create_forem_categories migration.
Note that the seeds only run in the installer if the migrations are also run.

db/seeds.rb
browning, line 1: the constantize gets called automatically when you use user_class in lib/forem.rb right now

radar, line 1: Ah, you're right.

radar:
Merged, thank you! :)

PR #127, opened 2012-01-09
Add atom feed to forums

jgadbois:


jgadbois:
Icon is from here: http://www.mozilla.org/foundation/feed-icon-guidelines/

jgadbois:
Any thoughts on this?  Happy to change it if you have some concerns before pulling

knewter:
@jgadbois my only concern in general is it'd be nice if there was a test to go along with it / verify we don't break this feature down the road.  Can you put one of those together?

jgadbois:
Just an integration test I'm guessing?

knewter:
Yeah, and it can be super simple as far as I care.  Just 'hey here's a forum/topic/post, and they totally show up in the rss feed when I hit the url.'

jgadbois:
Ok, done

knewter:
+1, I tested it and it both passes and merges cleanly.  @radar @parndt merge?  Thanks, you're great @jgadbois

parndt:
I'm not sure Atom is a technology that is useful in a Forum.

Just kidding! This sounds great to me, with one caveat. It needs something in the readme that explains that it's there and that you can place an auto discovery tag in your `<head>` to make it magic.

jgadbois:
Haha - had me nervous there!  Let me know if there are any other features you'd like help on.  I'll update the readme.

jgadbois:
Hmm...actually.  How would this work - the atom feed is per forum, so it's only valid on a forum page.  Is the auto_discovery_link_tag method smart enough to know this or would there need to be some conditional logic in the head?  

If the second, maybe we should add a helper method?

parndt:
Yes, a helper sounds perfect to me!

jgadbois:
It looks like since the engine has an isolated namespace , then helpers aren't available to the main application unless the helper is manually included in the app ApplicationController.  Is there a better solution, or should that just be part of the setup instructions?

parndt:
We should surely be able to have something like: 

forem.atom_auto_discovery_link_tag

@radar @knewter what is your feeling on a "forem" helper?

jgadbois:
Ok, I was just working on a helper method called forem_auto_discovery_link_tag that outputted the auto discovery tag when on forums#show

knewter:
So I would suggest you just make the helper named `auto_discovery_link_tag`.  Then, if the main app mounts forem at /forem, they can call that helper method with `forem.auto_discovery_link_tag`

parndt:
Seems crazy to override the real method name Rails provides..

knewter:
hilariously I forgot about that.  Ignore the crazy person, move along :)

jgadbois:
What if we just instructed them to put a yield(:forem_head) in their layout and then we did a content_for with the auto discover tag in the appropriate views.

radar:
You know, you could probably get the installer to insert it itself. http://rubydoc.info/github/wycats/thor/master/Thor/Actions#insert_into_file-instance_method

parndt:
Inserting it seems relatively brittle, how do we ensure we get the right layout?

jgadbois:
I agree, it would be hard to find the correct files.  This approach might still be easier than opening up a helper to the main app.

parndt:
I'd merge this request if you were to please add a helper named: `forem.atom_auto_discovery_link_tag` and document its usage in the readme.

Thanks, you're wonderful :)

parndt:
Or, if you (or Monique) think that name is dumb then you could rename the helper to something clearer :)

jgadbois:
How do I go about adding that method?  I can add a method to a helper, but I'm not sure how to make it accessible via `forem.atom_auto_discovery_link_tag`

knewter:
@radar @parndt I don't know the answer to this.  Le help?  So close on this one...

radar:
Sorry, was up a mountain (with @parndt and ~60 other nerds).

You can't make it available through the `forem` method, as that's only supposed to be used as a routing proxy. Instead, just define it as `forem_atom_auto_discovery_link_tag` and use it where appropriate.

jgadbois:
@radar - sounds like fun!  What is the best way to do make this method available as the engine's namespace is isolated (thus helpers aren't available to the main app) ?

radar:
I don't know :( Perhaps @parndt knows? Maybe he's done it with Refinery?

parndt:
You have to add the helper to the application's helper.. here's how Refinery does it:

https://github.com/resolve/refinerycms/blob/master/core/lib/refinery/core/engine.rb#L28

jgadbois:
I added the forem auto discovery helper method.

parndt:
```
Failures:

  1) topics viewing a topic should have an autodiscover link tag
     Failure/Error: assert page.has_selector?("link[title='ATOM']")
     MiniTest::Assertion:
       Failed assertion, no message given.
     # (eval):2:in `assert'
     # ./spec/requests/topics_spec.rb:153:in `block (3 levels) in <top (required)>'

Finished in 10.92 seconds
132 examples, 1 failure

Failed examples:

rspec ./spec/requests/topics_spec.rb:151 # topics viewing a topic should have an autodiscover link tag
```

jgadbois:
Strange the spec passed for me

parndt:
I'm running Ruby 1.9.3, what are you running?

jgadbois:
Same.  I'm afk, but I'll double check tonight. I don't think I used minitest though.

jgadbois:
@parndt This is passing for me

```

john@ubuntu:~/projects/github/forem$ bundle exec rspec spec/requests/
/home/john/.rbenv/versions/1.9.3-p0/lib/ruby/gems/1.9.1/gems/ffi-1.0.9/lib/ffi/platform.rb:27: Use RbConfig instead of obsolete and deprecated Config.
.............................................................................

Finished in 13.73 seconds
77 examples, 0 failures
john@ubuntu:~/projects/github/forem$ ruby -v
ruby 1.9.3p0 (2011-10-30 revision 33570) [i686-linux]
john@ubuntu:~/projects/github/forem$ git branch
* atom_feed
  ignore_dummy
  issue_137
  issue_150
  issue_153
  master
```

knewter:
tests pass for me as well...

knewter:
alright, tests pass for me on 2 different boxes.  Also, @parndt is crazy because why is minitest in his error.  Merging :)

parndt:
+1 I'm crazy

radar:
+1 @parndt is crazy.

jgadbois:
haha, thanks for merging this guys.

Issue #132, opened 2012-01-17
Category stats include hidden topics for non-admin users


mahzilla:
If there is a hidden topic in a forum, the Topics, Posts, and Views statistics for ForumsController#index and ForumsController#show are different if I view the forum as a non-admin user.

To replicate this issue, create a new forum, create a new topic, and then hide the topic.  You will see that the as a non-admin user, the newly create forum in ForumsController#index shows that a topic exists, but ForumsController#show will display "There are no topics in this forum currently."

radar:
Without looking at the code I can confirm this is an issue (seriously). Would you be able to submit a patch to fix this? It should be in `app/views/forem/forums/_forum.html.erb`. It just needs to change to `forum.topics.visible.count` or something like that. Oh, and the posts will also need to be changed.

mahzilla:
It was pretty easy to change the statistics in the view for topic count by adding `forem_admin? ? forum.topics.count : forum.topics.visible.count`, but I'm not quite sure how to change the models for posts or views to exclude the count for hidden parent topics.

radar:
To do that you could add a new scope to both the `Forem::Post` and the `Forem::View` model also called `visible` that did something like this for `Forem::Post`:

``` ruby
def self.visible
  joins(:topic).where("topics.hidden" => false)
end
```

And a small change for `Forem::View`

``` ruby
def self.visible
  joins(:post => :topic).where("topics.hidden" => false)
end
```

With those two scopes you could implement the same thing as you've just done for the topics count.

suciuvlad:
is this still open ?

parndt:
@suciuvlad  it's still open but I have no idea if it's still an issue... @radar ?

radar:
I believe this issue to be fixed now.

PR #134, opened 2012-01-19
Addresses Issue #132

mahzilla:
Hidden topics will only be included in the count if the user is an
admin when viewing the views/forem/forums/_forum.html.erb template

### DIFF
--- a/app/models/forem/post.rb
+++ b/app/models/forem/post.rb
@@ -10,6 +10,7 @@ module Forem
 
     delegate :forum, :to => :topic
 
+    scope :visible, joins(:topic).where("forem_topics.hidden" => false)
     scope :by_created_at, order("created_at asc")
 
     validates :text, :presence => true
--- a/app/models/forem/view.rb
+++ b/app/models/forem/view.rb
@@ -3,6 +3,8 @@ module Forem
     belongs_to :topic
     belongs_to :user, :class_name => Forem.user_class.to_s
 
+    scope :visible, joins(:topic).where("forem_topics.hidden" => false)
+    
     validates :topic_id, :presence => true
   end
 end
--- a/app/views/forem/forums/_forum.html.erb
+++ b/app/views/forem/forums/_forum.html.erb
@@ -12,8 +12,8 @@
           <% end %>
         </span>
     </td>
-    <td class="topics-count"><%= forum.topics.count %></td>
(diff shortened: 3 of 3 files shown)


knewter:
@mahzilla the code looks legit, but I'd love to see a new test related to it.  If you add a test, I'll gladly +1 this as it's needed.

knewter:
@mahzilla I should also say, thanks for the contribution.  It's awesome that you pushed this :)

app/views/forem/forums/_forum.html.erb
```diff
@@ -12,8 +12,8 @@
           <% end %>
         </span>
     </td>
-    <td class="topics-count"><%= forum.topics.count %></td>
-    <td class="posts-count"><%= forum.posts.count %></td>
-    <td class="views-count"><%= forum.views.sum(:count) %></td>
+    <td class="topics-count"><%= forem_admin? ? forum.topics.count : forum.topics.visible.count %></td>
+    <td class="posts-count"><%= forem_admin? ? forum.posts.count : forum.posts.visible.count %></td>
+    <td class="views-count"><%= forem_admin? ? forum.views.sum(:count) : forum.views.visible.sum(:count) %></td>
```
parndt, line 17: We should probably use a helper here, for each line.

parndt:
I made this very hard to merge with my recent change 326327a9b4e6c76a783f5df22c6972d6ccedb836 sorry.

app/models/forem/post.rb
```diff
@@ -10,6 +10,7 @@ class Post < ActiveRecord::Base
 
     delegate :forum, :to => :topic
 
+    scope :visible, joins(:topic).where("forem_topics.hidden" => false)
```
parndt, line 13: Can you please change this to a method like `by_created_at` now is since 326327a9b4e6c76a783f5df22c6972d6ccedb836? :)

app/models/forem/view.rb
```diff
@@ -3,6 +3,8 @@ class View < ActiveRecord::Base
     belongs_to :topic
     belongs_to :user, :class_name => Forem.user_class.to_s
 
+    scope :visible, joins(:topic).where("forem_topics.hidden" => false)
```
parndt, line 6: Can you please change this to a method per 326327a9b4e6c76a783f5df22c6972d6ccedb836?

PR #136, opened 2012-01-20
Change dependencies to rails 3.2.0

jgadbois:


parndt:
This makes forem unusable on anything less than 3.2.0. Could the dependency specify >= 3.1, < 3.3 ?

jgadbois:
I think getting rid of the .0 after 3.1 should work.

From the bundler site:

> ~> 2.1 is identical to >= 2.1 and < 3.0

parndt:
I do believe that'd work as apparently 3.2 is the last version before 4.0

radar:
@parndt I find that hard to believe given that 4.0 is supposed to be released in mid-2013. I wouldn't be surprised if they did a 3.3.

PR #140, opened 2012-02-03
Add forem to nested topic links.  Fixes #137

### LINKED ISSUE #137: Subscribe/Unsubscribe links in topics/show don't include forum
```
 <%= link_to t(".subscribe"), subscribe_forum_topic_path(@topic) %>
```

I think should be

```
 <%= link_to t(".subscribe"), subscribe_forum_topic_path(@forum, @topic) %>
```

jgadbois:


parndt:
How do you feel about writing a spec that ensures your expected behaviour remains and fails without your patch? :)

jgadbois:
For some reason the specs and test app work even without this change, so I'm not sure how to go about doing that.  There's an integration test that I added previously that tests these links.

parndt:
Very odd, I would have thought that this would lead to an improper route without the `@forum` attribute. Maybe we're not testing this page properly? Maybe both routes work? Weird!

jgadbois:
It didn't work for me in my production environment, but I manually tested on the test app in addition to my spec and it worked.

parndt:
Thanks for fixing this @jgadbois. :)

PR #144, opened 2012-02-08
User model avatars

codev:
I didn't want to use gravatar on a recent project for anonymity but wanted avatars from my current user class. So I've added the ability to set Forem.avatar_user_method to a method to be called on user to return an image url to use as an avatar. If you don't set avatar_user_method you keep the gravatar functionality.


parndt:
Odd, it's saying there are merge conflicts with this pull request. Can you please rebase against master?

Thanks!

codev:
Any better?

parndt:
Yeah, it now is clean.

parndt:
Or was, sorry.

codev:
Merged it with all the moderation fixes

radar:
Please extract changes pertinent to this pull request into another branch and re-submit a pull request. There are far too many commits in here with too many changes for my liking.

For instance, you have a commit that removes subscriber functionality and `lib/forem.rb` removes the `forem_autocomplete` attribute.

spec/dummy/db/migrate/20120208202625_add_custom_avatar_url_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddCustomAvatarUrlToUsers < ActiveRecord::Migration
```
radar, line 1: spec/dummy is re-generated with a call to `rake forem:dummy_app`, and so this migration would be nuked by that call. Recommended to move it into the dummy generator.

app/helpers/forem/posts_helper.rb
```diff
@@ -1,8 +1,16 @@
 module Forem
   module PostsHelper
     def avatar(user, options = {})
-      if (email = user.try(:email)).present?
-        image_tag(avatar_url(email, options), :alt => "Gravatar")
+      if Forem.avatar_user_method.nil?
```
parndt, line 4: if `avatar_user_method` just returned `:email` by default then you wouldn't need the conditional logic here at all :-)

parndt, line 4: By this, I mean that `avatar_url` would need to be modified to handle all cases automatically.

Issue #141, opened 2012-02-03
Migrations differ in db/migrate to spec/dummy/db/migrate


parndt:
@knewter @radar do you know why?

knewter:
My guess is because spec/dummy/db/migrate shouldn't be in the repo in the first place.  I have other engines where I've pushed them up, and it's always been a mistake then.

parndt:
Agree, I think we should refactor it out and make it a task instead.

radar:
Please Do Investigate

parndt:
Oh no he didn't

ugisozols:
Please see #149.

radar:
This is fixed in #149.

PR #149, opened 2012-02-15
Remove dummy app from spec folder and instead create rake task which wil...

ugisozols:
...l generate dummy app when needed.

Hello gentlemen!

I noticed #141 and decided to do something about it. I basically copy/pasted the code which is in Refinery CMS with some tweaks for the specs to pass.

Please review it and let me know if there's anything else needed.


parndt:
Looks good to me! @radar? @knewter?

knewter:
README needs updating...

also blah blah something about refinery, amirite @pardnt

parndt:
@knewter nice spelling ;) What do we need about Refinery?

knewter:
I was mocking me-on-other-ticket getting confused about which project I was talking about :)  _whoosh_

parndt:
@knewter :trollface:

benatkin:
I found one problem with this pull request. If I create a new rails app and add this to the Gemfile:

``` ruby
gem 'forem', :git => 'https://github.com/ugisozols/forem.git', :branch => 'dummy_app'
```

When I run `bundle install` and then `rails g -h` I get this:

```
Forem:
  forem:dummy
  forem:install
  forem:views
```

ugisozols:
I'm sorry but what's the problem?

benatkin:
It adds an extra generator, `forem:dummy`, which isn't relevant for rails apps that use forem. It's for forem development only. Refinery puts it in a separate folder, `testing/lib/generators`.

benatkin:
I made a [pull request](https://github.com/ugisozols/forem/pull/1) to your branch , @ugisozols, with changes to hide `forem:dummy` from `rails g -h` inside of an app that uses `forem`. I moved it to `spec/lib` instead of `spec/support` because rspec automatically pulls in everything in `spec/support` including templates and that caused the specs to fail.

ugisozols:
:+1: @benatkin. I agree with you that dummy_app rake task should be only available when developing forem. Thanks!

radar:
So you guys reckon this is cool to merge then?

ugisozols:
Specs pass - what's cooler than that? :)

radar:
ICE COLD.

jgadbois:
How do I regenerate the dummy app for development purposes?

ugisozols:
`bundle exec rake forem:dummy_app`

jgadbois:
yeah sorry, just saw the updated readme

jgadbois:
Should spec/dummy/ be added to the .gitignore file?

radar:
Yes.

jgadbois:
Ok, sorry I must not have been looking at the correct branch - my .gitignore didn't have it and git was prompting me to readd the dummy files.  I see that it was added in the initial commit.

stas:
I was looking for something like this and came up with a different solution, thought you might be interested to check it:

``` ruby
require 'rails/generators'
require 'rails/generators/rails/plugin_new/plugin_new_generator'

module MyEngine
  class DummyGenerator < Rails::Generators::PluginNewGenerator

    def self.default_source_root
      Rails::Generators::PluginNewGenerator.default_source_root
    end

    def do_nothing
    end

    alias :create_root :do_nothing
    alias :create_root_files :do_nothing
    alias :create_app_files :do_nothing
    alias :create_config_files :do_nothing
    alias :create_lib_files :do_nothing
    alias :create_public_stylesheets_files :do_nothing
    alias :create_javascript_files :do_nothing
    alias :create_script_files :do_nothing
    alias :update_gemfile :do_nothing
    alias :create_test_files :do_nothing
    alias :finish_template :do_nothing

  end
end
```

Where in Rakefile I just call the generator:

``` rake
task :setup do
  require 'rails'
  require 'myengine'
  require 'myengine/generators/dummy_generator'

  MyEngine::DummyGenerator.start(
    %w(. --quiet --force --skip-bundle --old-style-hash --dummy-path=spec/dummy)
  )
end
```

Thanks

radar:
I've been looking at using the `dummier` gem for things like this. I think we should switch over to using that. It provides its own Rake task for generating the dummy application as well as some very useful hooks.

stas:
Thanks for `dummier`, it seems nice.
Still It looks pretty tied to `Test::Unit`, so it will likely have to be patched first.

PR #154, opened 2012-02-20
Touch topics when posts are updated to enable ordering by post without a...

jgadbois:
... join

This solution will probably require a rake task or migration for existing installations.

### DIFF
--- a/app/models/forem/post.rb
+++ b/app/models/forem/post.rb
@@ -1,6 +1,6 @@
 module Forem
   class Post < ActiveRecord::Base
-    belongs_to :topic
+    belongs_to :topic, :touch => true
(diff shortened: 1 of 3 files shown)


radar:
Why would this require a rake task or migration for existing installations?

jgadbois:
Because in existing installations topic updated_at will be the same as topic created_at, causing the topics to be ordered by creation instead of by topics with the most recent posts.  Once another post was added it would move up to the top correctly.

radar:
Rightio, well use `Forem::Topic.find_each` inside that migration to fix it then. Gogo ;)

jgadbois:
@radar - added migration

PR #156, opened 2012-02-21
Moderation system

radar:
I'm currently using Forem for a Spree project (the Spree Forums), and we've decided that we're going to need a moderation system. 

We're currently switching from Google Groups which automatically moderates each person's first post, hiding it from the rest of the list and notifying the assigned moderators of there being a new moderated post. This is probably one of about three things that Google Groups does right. The moderators then can choose to "Deny", "Post", "Ignore" or "Spam" this post.

"Deny" means to completely hide the post from the list with an optional area to allow the moderator to explain to the user why their post was denied. Good for handling first-offenders.

"Post" makes the post appear on the mailing list, sending out notifications to people that the post has been created right then, when in actual fact it's just been moderated.

"Ignore" is like deny, without the reason. Good for hellbanning people, although I've personally never used it.

"Spam" marks this user account as spam, and bans them from posting to the group. Great for obviously spammy posts, but this is something that could be automated with something like acts_as_snook.

---

I would recommend that we follow this same path with Forem, with some tweaks.

All new posts and topics are assigned to **that forum's** moderation queue when they are created. The moderators of the system can then go through and choose whatever option they want to to deal with these posts.

While that's happening, the user who created the topic/post and the moderators of that forum should be able to see the topic/post, with a warning stating that it's currently sitting in the moderation queue.

Once it's out of the moderation queue, that user is then marked as "safe", with each subsequent topic and post skipping the whole moderation mumbo-jumbo and going right onto the forums.

---

Now, there's inevitably going to be a point where someone begins trolling or being an ass. This is where other users can report those people, clicking a link on the post to report it.

 In a moderator interface, a moderator should be able to see all the posts that have been reported, with the post or topic with the highest number of reports being first on that list. The moderator can then choose to perform moderator-type actions on it, like "Hide" or "Lock" (which are currently available actions), or "Ignore", which simply ignores all further reports from this topic/post.

Moderators should also be able to ignore reports from a specific user, if that user is being troublesome, although this is really a nice-to-have.

---

This pull request contains the beginnings of the moderation system as I've just described. Patches welcome ;)


gabrielbaez:
What if the website admin does not wish to moderate all posts or only wants this feature for a specific   forum (ie support) and not all?  Can this be turned  on/off ?

radar:
They can choose to assign no moderator groups to a forum meaning that nobody (bar the admins) will have moderator rights on that forum.

gabrielbaez:
ah ok thanks,  I'm surprise you are awake! I though you were on the other side of the world!

radar:
I usually am, but at this current point in time I am in DC (Bethesda) working out of the Spree offices. I'll be in Americaland until the end of April most probably.

PR #162, opened 2012-02-29
ApproveAllTopicsAndPosts

glebm:
`rake spec' green


parndt:
So we have to manage updated_at manually? I suppose that makes sense.

@radar?

radar:
Why are you setting the updated_at field? I am not sure I understand that part.

parndt:
I'm guessing because the hook that normally would is being bypassed?

radar:
What effect does setting it have though other to indicate when the topic was approved at? Afaik we don't use it anywhere in forem.

glebm:
@parndt Yes, but of course we don't want that hook here (because the existing sort order on updated at would break?)
@radar Updated!

radar:
Update all is definitely a more sensible choice. Thanks :)

PR #163, opened 2012-03-02
More detailed view tracking + visual indication of new topics

jgadbois:


jgadbois:
Still need some specs for overall view tracking of forums and topics.  Also specs for register_view_by(user) to make sure that is only updating timestamps when appropriate.

jgadbois:
@radar - Added specs - think this is good to go.  Please let me know what you think.

radar:
Yup, seems good. @parndt, go for it!

parndt:
Did it live

jgadbois:
Thanks guys!

PR #170, opened 2012-03-07
French localization now available.

dlepage:
One more language...

### DIFF
--- a/config/locales/fr.yml
+++ b/config/locales/fr.yml
@@ -9,14 +9,14 @@ en:
         zero: Aucun message
         one: 1 message
         other: "%{count} messages"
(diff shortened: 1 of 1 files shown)


config/locales/fr.yml
```diff
@@ -0,0 +1,197 @@
+fr:
```
parndt, line 1: this should be `fr:`

parndt, line 1: which I'm guessing you just noticed at the same time!

dlepage:
fixed ! Sorry

parndt:
Merged in cd49c29 thank you!

config/locales/fr.yml
dlepage, line 1: You bet ! :)

PR #174, opened 2012-03-12
Add per page configuration and redirect to last page when posting replies

jgadbois:
Currently everything redirects back to the first page.


app/controllers/forem/posts_controller.rb
```diff
@@ -78,5 +78,11 @@ def block_spammers
         redirect_to :back
       end
     end
+
+    def last_page
+      if @topic.posts && Forem.per_page
+        (@topic.posts.size.to_f / Forem.per_page.to_f).ceil
```
radar: Please use `count` here as it will do an SQL count, where `size` will load all posts objects into memory and then do `size` on that array.

PR #177, opened 2012-03-16
Fixes some nil references

eivindhagen:
In my forum I allowed people to post without signing up for a user account, so those posts were created with User objects that did not have an id set (they were new_record?)

I found a couple of places where adding nil checks avoids crashes.

### DIFF
--- a/app/models/forem/topic.rb
+++ b/app/models/forem/topic.rb
@@ -149,7 +149,9 @@ module Forem
       if state_changed?
         first_post = self.posts.by_created_at.first
         first_post.approve! unless first_post.approved?
-        self.user.update_attribute(:forem_state, 'approved') if self.user.forem_state != 'approved'
+        if self.user
+          self.user.update_attribute(:forem_state, 'approved') if self.user.forem_state != 'approved'
+        end
       end
     end
   end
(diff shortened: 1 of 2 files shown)


parndt:
I'm not certain how much I am a fan of this concept.. it makes the code quite defensive.

app/models/forem/topic.rb
```diff
@@ -149,7 +149,9 @@ def approve_user_and_posts
       if state_changed?
         first_post = self.posts.by_created_at.first
         first_post.approve! unless first_post.approved?
-        self.user.update_attribute(:forem_state, 'approved') if self.user.forem_state != 'approved'
+        if self.user
+          self.user.update_attribute(:forem_state, 'approved') if self.user.forem_state != 'approved'
```
parndt, line 153: can simply be:

``` ruby
self.user.update_attribute(:forem_state, 'approved') if self.user && self.user.forem_state != 'approved'
```

app/views/forem/topics/show.html.erb
```diff
@@ -5,7 +5,7 @@
     <% if @topic.can_be_replied_to? && can?(:reply, @topic) %>
       <%= link_to t(".reply"), new_topic_post_path(@topic) %>
     <% end %>
-    <% if @topic.user == forem_user || forem_admin? %>
+    <% if (@topic.user and @topic.user == forem_user) || forem_admin? %>
```
parndt, line 8: You're changing the precedence by using `and` (note we don't use `or` we use `||`)

``` ruby
@topic.user && @topic.user == forem_user
```

radar:
I agree with Phil, this does make the code more defensive. Is it that all people can post without using an account? If so, why not override `Forem::Topic` to return a `User.new` object all the time, rather than `nil`?

PR #187, opened 2012-03-30
Fix setting topic.last_post_at value when creating a new post.

yan-hoose:
topic.last_post_at was set but topic was not saved. I changed it to use update_attribute.


radar:
Thank you!

PR #190, opened 2012-04-11
User model avatars

codev:
Rebased - Includes radar's spec fix - see inline for parndt's comments on the last commit.

I didn't want to use gravatar on a recent project for anonymity but wanted avatars from my current user class. So I've added the ability to set Forem.avatar_user_method to a method to be called on user to return an image url to use as an avatar. If you don't set avatar_user_method you keep the gravatar functionality.

### DIFF
--- a/app/helpers/forem/posts_helper.rb
+++ b/app/helpers/forem/posts_helper.rb
@@ -1,8 +1,15 @@
 module Forem
   module PostsHelper
     def avatar(user, options = {})
-      if (email = user.try(:email)).present?
-        image_tag(avatar_url(email, options), :alt => "Gravatar")
+      if Forem.avatar_user_method.nil?
+        if (email = user.try(:email)).present?
+          image_tag(avatar_url(email, options), :alt => "Gravatar")
+        end
+      else
+        # Try to use the user's custom avatar method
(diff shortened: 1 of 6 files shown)


app/helpers/forem/posts_helper.rb
```diff
@@ -1,8 +1,15 @@
 module Forem
   module PostsHelper
     def avatar(user, options = {})
-      if (email = user.try(:email)).present?
-        image_tag(avatar_url(email, options), :alt => "Gravatar")
+      if Forem.avatar_user_method.nil?
```
codev, line 4: parndt said "if avatar_user_method just returned :email by default then you wouldn't need the conditional logic here at all :-) By this, I mean that avatar_url would need to be modified to handle all cases automatically."

That is true but then the conditional logic would have to go into avatar_url to decide if the string is an email address or url and call gravatar or return the url. That might involve adding another setting. I'm happy to change the avatar_url to put the logic in there if you want.

parndt:
Thanks!

parndt:
Plus 25b2dacd83394c058001623449b6cde8a92edcd9 and 5dcbfbe66c3b1971c553d31aeacd735fd361b9c9 and 465eb99f59437de4f2ced907dfb572c2ad9c4d29

PR #192, opened 2012-04-11
Url slugs

glebm:
Please take a look at this draft implementation for URL slugs

Provides a default of "#{name}-#{id}", and allows to override it with pretty much anything

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .bundle/
+.idea/
 pkg/
 *.sqlite3
 *.log
--- a/app/controllers/forem/admin/categories_controller.rb
+++ b/app/controllers/forem/admin/categories_controller.rb
@@ -43,7 +43,7 @@ module Forem
 
       private
         def find_category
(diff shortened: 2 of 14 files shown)


parndt:
Couldn't we just use [friendly_id](/norman/friendly_id#readme) for this?

glebm:
Nice, didn't know about friendly_id. Looks great!
Although I am always a bit suspicious towards gems that override ActiveRecord::Base.find

parndt:
It only overrides on the models where you want it to by specifying (for example): 

class MyModel < ActiveRecord::Base
  extend FriendlyId

  friendly_id :title, :use => :slugged
end

I've been using it for years and it works well from my perspective. Of course I've been added as a collaborator to it to declare my bias ;-)

radar:
+1 for friendly_id

glebm:
OK then, I'll implement friendly_id :)

parndt:
@glebm thanks :-)

PR #193, opened 2012-04-11
Installation Error - rake task fails on categories

rodleg:
...ue

Note:  I'm a newbie and this is my first pull request.  

I ran across issue 189 while trying to install forem.  The migration fails when trying to update the categories table with "General".  I was able to resolve this issue by setting attr_accessible :name in the category.rb model.  

Thanks-
Rodney


radar:
This will indeed fix just the Category model, but there's other issues as well.

PR #196, opened 2012-04-14
Added column to user table: forem_auto_subscribe. True if user replies s...

jeremyong:
...hould automatically subscribe users, false otherwise. Set to false by default but can be changed in another migration


jeremyong:
Sorry this is a bad commit. Ignore this pull request please.

PR #203, opened 2012-04-23
Don't hard code users table for auto subscription option

jgadbois:


parndt:
Looks good to me cc @radar @knewter

jgadbois:
I was also wondering if setting up a store for this (and future?) user settings would make sense so every new setting doesn't need another column.

jeremyong:
Ah good catch. How would you suggest storing forem user settings? As a JSON blob?

jgadbois:
I was thinking of using ActiveRecord::Store which serializes to json I believe.

parndt:
What about ActiveSupport::Concern

jgadbois:
Where would we use ActiveSupport::Concern?  If we wanted to add settings to multiple models?

parndt:
Wait - what sort of user are you talking about? I think of users like developers sometimes which I think is wrong in this case? ;-)

jgadbois:
Sorry, I was talking about forem end-users.  I could see them having more options in the future (and potentially a front-end to edit these options) in addition to this "auto-subscribe" option.

jeremyong:
ActiveRecord::Store looks good. Should we do it immediately or wait for people to figure out what other sorts of settings users might need?

Another setting I thought of is an email setting (do you want to receive emails from administrators? etc)

jgadbois:
Let's wait and see what the forem team thinks.  Not a huge deal, just trying to cut down on the number of migrations needed going forward

knewter:
+1 on "we need a more generic settings system"

To use rails properly, this is ideal here: http://api.rubyonrails.org/classes/ActiveRecord/Store.html

Amirite?

radar:
Urrite

PR #205, opened 2012-04-24
Turn on parameter whitelisting and get specs to pass

### LINKED ISSUE #204: Mass assignment protection fails tests
``` ruby
"forum"=>{"category_id"=>"1", "title"=>"Buying Land", "description"=>"This is the section you will buy if you want one.", "moderator_ids"=>["3", ""]}, "commit"=>"Create Forum"}
```

ERROR

``` ruby
ActiveModel::MassAssignmentSecurity::Error (Can't mass-assign protected attributes: category_id, title, description, moderator_ids):
```

The whole thing needs testing for mass assignment.

jgadbois:
Fixes #204

### DIFF
--- a/spec/lib/generators/forem/dummy/dummy_generator.rb
+++ b/spec/lib/generators/forem/dummy/dummy_generator.rb
@@ -56,6 +56,11 @@ module Forem
       inject_into_file "#{dummy_path}/config/application.rb",
                   "\nrequire 'kaminari'\n",
                   :before => "module Dummy"
+
+      inject_into_file "#{dummy_path}/config/application.rb",
+                  "\n  config.active_record.whitelist_attributes = true\n",
+                  :before => "end\nend\n",
+                  :verbose => false
     end
 
     protected
--- a/app/models/forem/forum.rb
+++ b/app/models/forem/forum.rb
@@ -11,6 +11,8 @@ module Forem
 
     validates :category, :title, :description, :presence => true
 
+    attr_accessible :category_id, :title, :description, :moderator_ids
+
     def last_post_for(forem_user)
       forem_user && forem_user.forem_admin? || moderator?(forem_user) ? posts.last : last_visible_post
     end
--- a/app/models/forem/group.rb
+++ b/app/models/forem/group.rb
@@ -5,6 +5,8 @@ module Forem
     has_many :memberships
     has_many :members, :through => :memberships, :class_name => Forem.user_class.to_s
 
+    attr_accessible :name
+
     def to_s
       name
     end
--- a/app/models/forem/membership.rb
+++ b/app/models/forem/membership.rb
@@ -2,5 +2,7 @@ module Forem
   class Membership < ActiveRecord::Base
     belongs_to :group
     belongs_to :member, :class_name => Forem.user_class.to_s
+
+    attr_accessible :member_id, :group_id
   end
 end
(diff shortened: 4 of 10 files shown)


knewter:
Merged this and pushed.  Pushed some tweaks of my own that fixed issues that cropped up with Gemfile.lock update

safarista:
Just had a look and this comes up. This pull seams to break `undefined method forem_auto_subscribe?' for #<User:0x000001285548e8>`

``` ruby
{"utf8"=>"✓",
 "authenticity_token"=>"3tTj+Kj3qZi4jhISeEG7aaT7QOlb/1CgX1Q9n3LrnyQ=",
 "topic"=>{"subject"=>"wasa",
 "posts_attributes"=>{"0"=>{"text"=>"mamamama"}}},
 "commit"=>"Create Topic",
 "forum_id"=>"4"}
```

Will have a dig but if anyone has time before tomorrow go ahead and dig. Thanks

knewter:
we need a test case that shows this.  Sorry.  Anyone else seeing it?  @jgadbois?

jgadbois:
I didn't see it - maybe your User model just doesn't have the new forem_auto_subscribe field.  

You will potentially need to add it with your own migration because the migration to add it is dynamically generated during install, but not if you just copy over new migrations for upgrade.

safarista:
ah good eye. its true in my users table there is no such field. 

I'll do a migration now.

safarista:
@jgadbois where is the migration generator for this i want to have a look.

jgadbois:
lib/generators/forem/install/templates/forem_auto_subscribe_migration.rb

safarista:
@jgadbois thanks, That fixed it. 
For those like me on a steep learning curve.
`add_column :users, :forem_auto_subscribe, :boolean, default: false`

jgadbois:
Yeah unless your Forem user class isn't 'User'.  If it is something else then you will need to change :users.

spec/lib/generators/forem/dummy/dummy_generator.rb
```diff
@@ -56,6 +56,11 @@ def test_dummy_config
       inject_into_file "#{dummy_path}/config/application.rb",
                   "\nrequire 'kaminari'\n",
                   :before => "module Dummy"
+
+      inject_into_file "#{dummy_path}/config/application.rb",
+                  "\n  config.active_record.whitelist_attributes = true\n",
```
radar, line 61: Strongly strongly strongly disagree.

jgadbois, line 61: Suggestions?  You disagree with approach or turning whitelist_attributes on?

radar, line 61: Disagree with turning this option on. If you turn this option on there's no way to know for certain that we have all attributes mass-assignable as required.

I'm about to head out to lunch but will look at this again when I come back.

jgadbois, line 61: K, I was just turning it on so that things actually break when running the specs.  Otherwise everything can be mass assigned no problem (unless attr_accessible is called).

radar, line 61: I've now removed this with @35bd6fa. All the tests still pass. Nice work!

PR #209, opened 2012-04-25
Forem.base_path does not have trailing slash

jgadbois:


parndt:
Weird that we don't just use named routes

jgadbois:
In the js file?

parndt:
Yeah we could consider .js.erb cc @radar @knewter

radar:
I wasn't aware that routes would work in an asset file.

jgadbois:
If you can that would work fine for the autocomplete path, but not for the groups path.

PR #219, opened 2012-05-11
Fix loop

tonybruess:
Previously if you had more than one category, it would loop through the forums from the first category for all the categories.

This fixes that :P

### DIFF
--- a/app/views/forem/admin/forums/index.html.erb
+++ b/app/views/forem/admin/forums/index.html.erb
@@ -16,7 +16,7 @@
       </tr>
     </thead>
     <tbody>
-      <% @forums.each do |forum| %>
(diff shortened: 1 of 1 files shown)


parndt:
Thank you. Any chance of a spec that fails before this change and passes after it?

tonybruess:
I've done zip with specs, so, I wouldn't be of any help on that ;(

safarista:
Actually this is more of a typo if you ask me. I did not run a test but the logic in the loop will tell you. I forgot to push a fix a while ago when i discovered it. I would just do a merge @parndt

parndt:
Sure

parndt:
The reason I ask for specs is so that the functionality does not get broken in future :-)

PR #225, opened 2012-05-16
Use h1 tag for highest level forums title heading

juozasg:
This is a very simple tweak that uses h1 for the highest level title ("Forums") for forums index view. This makes it easier to style as a part refinery app (which uses h1 tags for highest level titles by default). Also, it's higher level than category headings, which are h2.


parndt:
+1

PR #226, opened 2012-05-16
Updates to radar/1-0-stable

juozasg:
I've cherry picked few commits from master and added some of my own commits on top of your 1-0-stable branch. There are no major changes, mainly it involves moving some dependencies (kaminary, forem-redcarpet) into development gemspec. A small (non-active) bug in post.rb. There is one major dependency change (from state_machine to workflow) which I included to make other cherry picks from master possible without any patching.

I've also cherry-picked minor avatar changes to allow people to set Forem.avatar_user_method and use in 1-0-stable.

Also picked something from master to make the tests pass in 1.8.7 and updated the Gemfile.lock with new dependencies.

The main reason for all these changes is to have a clean 1-0 branch that is ready to integrate with refinery-cms.

I'm not sure if that matches your development process at all, but hope you pick these into 1-0-stable. I'm working on forums/intranet site based on forem and refinery and would like to be able to build on a stable foundation as I imagine others would like as well.

All specs should pass

Thank you,
Juozas


.travis.yml
```diff
@@ -3,3 +3,4 @@ rvm:
   - 1.8.7
   - 1.9.2
   - 1.9.3
+  - ree
```
parndt, line 6: why ree? It's considered end of life and is compatible with 1.8.7

parndt, line 6: I guess this probably came from master, I've just removed that in 9e15e572fba5afb0f7e4e011025aab1e9da863c5

juozasg, line 6: yep, just came from master

juozasg:
i've also learned that some moderation stuff is little broken in 1-0-stable. This moderation fix commit should be included in radar/1-0-stable 5915732423747c36027af26671ccada0eec91b5a It doesn't cherry-pick cleanly, so there might be some other changes in between. I'm going to look at it

juozasg:
Looking at this, there are tons of moderation fixes and tweaks that go on top of radar/1-0-stable so I'm not sure what the roadmap for 1.0 is? If 1.0 should include moderation it might make more sense just to stabilize master instead of incrementing 1-0-stable with patches from master? Lemme know what you think.

parndt:
I don't think 1-0-stable should include any new features just bugfixes.

parndt:
Please use the second approach of incrementing 1-0-stable with patches from master :-) thanks

PR #227, opened 2012-05-17
Show the post the user is replying to

juozasg:
When replying to a  particular post (pressing "reply" instead of "quote") we should see a the entire post we are replying to on the new post page. When replying to a topic, the topic subject is shown. Includes one new spec.

If this is included in master it will require updating forem themes also to add "clear:both" style to new_post form because it goes directly under the markup of the post we are replying to. That markup uses float style which breaks the new post form layout.

### DIFF
--- a/app/controllers/forem/posts_controller.rb
+++ b/app/controllers/forem/posts_controller.rb
@@ -7,9 +7,10 @@ module Forem
     def new
       authorize! :reply, @topic
       @post = @topic.posts.build
+      @reply_to_post = @topic.posts.find_by_id(params[:reply_to_id])
+
       if params[:quote]
-        @reply_to = @topic.posts.find(params[:reply_to_id])
-        @post.text = view_context.forem_quote(@reply_to.text)
+        @post.text = view_context.forem_quote(@reply_to_post.text)
       end
     end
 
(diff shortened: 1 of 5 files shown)


parndt:
Thanks! How would you like to make screenshots of this so that we can see what's changing?

juozasg:
This also requires updating the twist theme: https://github.com/radar/forem-theme-twist/pull/1

juozasg:
When you go and view a topic, each post has a reply button on it. When you click the reply button on a post that has body content of "reply no. 15" this is what you get: http://img69.imageshack.us/img69/2975/foremshowreplytopostscr.png

radar:
Is it necessary to create a different view for the reply post? Why can't you just use the post partial that already exists?

juozasg:
I've considered both and created a new one, because using the existing one would require adding more parameters to the partials (for example a variable to disable actions). The post partial is pretty complex already imo so I didn't want to add to it. Alternatively, I could factor out some parts of the post partial (like actions, moderation stuff, etc.) into other partials - then if I added a 'preview' or 'reply_to_post' parameter to the post partial the logic would still be less complex. Let me know what you prefer.

radar:
Ok, in that case then having a second template is probably a better idea.

PR #228, opened 2012-05-17
Pagination view tweaks

juozasg:
Few simple tweaks to how pagination widget is rendered. 
- list of topics and list of posts use exact same pagination view logic
- pagination widget is rendered above and below the list for improved usablity

### DIFF
--- a/app/helpers/forem/application_helper.rb
+++ b/app/helpers/forem/application_helper.rb
@@ -18,9 +18,9 @@ module Forem
 
     def forem_pages_widget(collection)
       if collection.num_pages > 1
-        ("<div class='pages'>\n" +
-        t('forem.common.pages') + ':' + forem_paginate(collection) + "\n" +
-        "</div>").html_safe
+        content_tag :div, :class => 'pages' do
+          (t('forem.common.pages') + ':' + forem_paginate(collection)).html_safe
+        end
       end
     end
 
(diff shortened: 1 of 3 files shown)


juozasg:
Fails on ruby 1.8.7 with 

/home/vagrant/.rvm/gems/ruby-1.8.7-p358/gems/rubypython-0.5.3/lib/rubypython/pyobject.rb:14: uninitialized constant RubyPython::PyObject::FFI (NameError)

Not if this is related to this commit in any way.

juozasg:
This also requires updating the twist theme: https://github.com/radar/forem-theme-twist/pull/1

app/helpers/forem/application_helper.rb
```diff
@@ -16,6 +16,14 @@ def forem_markdown(text, *options)
       forem_format(text)
     end
 
+    def forem_pages_widget(collection)
+      if collection.num_pages > 1
+        ("<div class='pages'>\n" +
```
radar: Not sure I like the use of hard-coded HTML here. Would a `content_tag` not be better?

radar:
Other than that one note, this pull request looks good to me. Fix that up and I'll merge it in.

juozasg:
updated with new commit to use content_tag helper

radar:
Looks good, thanks.

PR #231, opened 2012-05-23
Use Forem.formatter to sanitize input if it has sanitize method. Needed for HTMl/WYSIWYG formatting.

juozasg:
Very basic change: if a formatter has a sanitize method, use it. With a simple spec.

### DIFF
--- a/spec/helpers/formatting_helper_spec.rb
+++ b/spec/helpers/formatting_helper_spec.rb
@@ -63,6 +63,8 @@ describe Forem::FormattingHelper do
       describe "uses formatter quoting method if exists" do
          subject { helper.as_quoted_text(raw_html) }
          before { Forem.formatter.stub('respond_to?').with(:blockquote).and_return(true) }
+         before { Forem.formatter.stub('respond_to?').with(:sanitize).and_return(false) }
+
          before { Forem.formatter.stub(:blockquote).and_return("> #{markdown}") }
 
          it "quotes the original content" do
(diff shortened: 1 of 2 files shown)


juozasg:
Also created a new formatter for raw HTML. If you merge this commit into master, you might also want to update the wiki page listing formatters to include this one: https://github.com/juozasg/forem-html_formatter

rebasing
```diff
@@ -0,0 +1 @@
+../forem-rebasing
```
parndt: what is this?

spec/helpers/formatting_helper_spec.rb
```diff
@@ -70,6 +72,17 @@
          end
          it {should be_html_safe}
        end
+
+      describe "uses formatter sanitize method if exists" do
+        subject { helper.as_formatted_html(markdown) }
+
+        before { Forem.formatter.stub('respond_to?').with(:blockquote).and_return(false) }
+        before { Forem.formatter.stub('respond_to?').with(:sanitize).and_return(true) }
+        before { Forem.formatter.stub(:sanitize).and_return("sanitized it") }
```
parndt: 3 before methods for 1 test? What's going on here?

rebasing
juozasg: oi, this should have not been checked in. lemme fix this

juozasg: forced update since this really shouldn't be in the history

spec/helpers/formatting_helper_spec.rb
juozasg: just following the format above that uses several before statements for one test. i will fix this

parndt:
Merged into 020eb27 thank you!

PR #235, opened 2012-05-30
Can't view topic if current_viewed_at not set

artem-mindrov:
I have a number of topics created between the two migrations `add_forem_views` and `add_forem_view_fields`, so `viewed_at` fields for them are nil. Whenever I try to view one of these topics, I get a 'no method < for nil' from Viewable.

### DIFF
--- a/app/models/forem/concerns/viewable.rb
+++ b/app/models/forem/concerns/viewable.rb
@@ -22,6 +22,8 @@ module Forem
         increment!(:views_count)
 
         # update current viewed at if more than 15 minutes ago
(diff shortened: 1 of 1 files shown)


radar:
You didn't test this, did you?

artem-mindrov:
Yeah sorry, I was in a haste. Early morning :(

radar:
No worries, easy fix.

artem-mindrov:
I'm also wondering if the `add_forem_view_fields` migration should set the timestamps and add a non-null constraint on them?

PR #238, opened 2012-06-08
bug fix: moderation > work when option not selected

joxxoxo:
bug description:
when forum admin clicks "Moderate" without selecting 'Approve' or 'Spam', then
for topic: 500 error occurs (params[:topic] is nil)
for posts: nothing happens but success message is shown

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ tmp
 *.sw*
 spec/dummy
 *.rbc
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,3 @@ tmp
 *.sw*
 spec/dummy
 *.rbc
-.idea
--- a/app/models/forem/post.rb
+++ b/app/models/forem/post.rb
@@ -77,8 +77,10 @@ module Forem
       def moderate!(posts)
         posts.each do |post_id, moderation|
           # We use find_by_id here just in case a post has been deleted.
-          post = Post.find_by_id(post_id)
-          post.send("#{moderation[:moderation_option]}!") if post
+          if moderation[:moderation_option]
+            post = Post.find_by_id(post_id)
+            post.send("#{moderation[:moderation_option]}!") if post
+          end
         end
       end
     end
--- a/app/models/forem/post.rb
+++ b/app/models/forem/post.rb
@@ -77,9 +77,8 @@ module Forem
       def moderate!(posts)
         posts.each do |post_id, moderation|
           # We use find_by_id here just in case a post has been deleted.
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -10,3 +10,4 @@ tmp
 *.sw*
 spec/dummy
 *.rbc
+.idea
```
parndt: should be in your ~/.gitignore

app/models/forem/post.rb
```diff
@@ -77,8 +77,10 @@ def topic_not_pending_review
       def moderate!(posts)
         posts.each do |post_id, moderation|
           # We use find_by_id here just in case a post has been deleted.
-          post = Post.find_by_id(post_id)
-          post.send("#{moderation[:moderation_option]}!") if post
+          if moderation[:moderation_option]
+            post = Post.find_by_id(post_id)
```
parndt: this could probably be refactored with @radar's blessing to:

``` ruby
if moderation[:moderation_option] && post = Post.find_by_id(post_id)
```

.gitignore
joxxoxo: do you mean i should remove this from here? what is the harm of its presence?

app/models/forem/post.rb
joxxoxo: great idea, thanks

.gitignore
parndt: Do you add it to every project you work on?

radar: I don't agree with .idea being added to .gitignore. Rather, you should be careful with what you add to your commit, adding changes using preferably the -p option during git add, so that changes like debug statements or IDE pollutant directories don't get added. 

## 

Ryan Bigg

radar: Also, if any pull request submitted an .idea directory with the contents normally found in one, I would reject it outright. 

With an explanation, of course. 

## 

Ryan Bigg

joxxoxo: ok, i've removed it already. 
and yes, it was usually added to project's .gitignore automaticaly(for example by refinerycms generator), by me or by other collaborators. 
thanks for sharing your point of view.
may be you can name any disadvantages of adding this to project's .gitignore?

radar: There are no disadvantages about adding to it... It's just I would prefer it kept short and clean.  

Thanks for removing that .idea directory! 

## 

Ryan Bigg

joxxoxo: thanks for explanation!

radar:
This issue has now been fixed. Thanks for letting us know about it.

parndt:
@radar you broke the build with 505f10c

PR #240, opened 2012-06-12
remove second definition for PostS#subscribe_replier

joxxoxo:
this method already defined few lines upper


parndt:
Seems legit; thanks!

PR #243, opened 2012-06-13
Removed protocol from gravatar url to allow https.

beefsack:
I run the forum on a server using https, and the gravatar images were giving security warnings.  I've made the urls protocol relative as per: http://paulirish.com/2010/the-protocol-relative-url/ which has removed the security warnings for me.


parndt:
Thanks!

Do you think you could squash it down to one commit?

beefsack:
Done!

parndt:
Legend!

PR #246, opened 2012-06-15
Updated ssl Gravatar links to use https://secure instead of https://www

beefsack:
Another update to a change I made recently.  I just discovered that Gravatar prefers https://secure than https://www, so I added code to conditionally add one or the other depending on request.ssl?

New test added to test SSL Gravatar URLs.


radar:
Merged, thanks!

PR #250, opened 2012-06-19
fixes duplicate migration in 241 and moves to using railties for migration installation

jzahnd:


radar:
Please fix the failure caused by this pull request.

db/migrate/20120616193448_add_forem_auto_subscribe.rb
```diff
@@ -0,0 +1,12 @@
+# This migration comes from forem (originally 20120616000935)
```
parndt, line 1: This shouldn't be present..  :-)

parndt:
@jzahnd  thanks very much!! I've merged your pull request into 6a46e7e3949b4694d6ba6728972ed1ab210106c7 and added f44662c to handle the superfluous comments.

Appreciate your input :) thanks again.

PR #254, opened 2012-06-28
Added moderation and subscription_mailer folders to view generator

mmcc:
When I was modifying views I noticed the moderation folder wasn't generated. Figured I might as well add the subscription_mailer views as well.


radar:
Thank you!

PR #256, opened 2012-07-05
Swedish localization

kl:


parndt:
Thanks for this! Do you think you could squash this into a single commit please?

PR #260, opened 2012-07-12
Config engine to load decorators

kunalchaudhari:
While extending forem classes we need to require decorators in main app's application.rb. Now with this code they will automatically required when engine get loaded.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ All of Forem’s business logic (models, controllers, helpers, etc) can easily b
 Standard practice for including such changes in your application or extension is to create a file within the relevant app/models or app/controllers directory with the original class name with _decorator appended.
 
 ### Adding a custom method to the Post model:
-    # app/models/forem/post_decorator.rb
+    # app/decorators/models/forem/post_decorator.rb
 
     Forem::Post.class_eval do
       def some_method
@@ -107,7 +107,7 @@ Standard practice for including such changes in your application or extension is
     end
 
 ### Adding a custom method to the PostsController:
-    # app/controllers/forem/posts_controller_decorator.rb
+    # app/decorators/controllers/forem/posts_controller_decorator.rb
 
     Forem::PostsController.class_eval do
       def some_action
--- a/README.md
+++ b/README.md
@@ -83,13 +83,39 @@ If you would like to add auto discovery links for the built in forum Atom feeds,
 
     <%= forem_atom_auto_discovery_link_tag %>
 
-## Customisation
+## Views Customisation
 
 If you want to customise Forem, you can copy over the views using the (Devise-inspired) `forem:views` generator:
 
     rails g forem:views
 
 You will then be able to edit the forem views inside the `app/views/forem` of your application. These views will take precedence over those in the engine.
+
+## Extending Classes
+
+All of Forem’s business logic (models, controllers, helpers, etc) can easily be extended / overridden to meet your exact requirements using standard Ruby idioms.
+
+Standard practice for including such changes in your application or extension is to create a file within the relevant app/models or app/controllers directory with the original class name with _decorator appended.
+
+### Adding a custom method to the Post model:
+    # app/models/post_decorator.rb
+
+    Forem::Post.class_eval do
+      def some_method
+        ...
+      end
+    end
+
+### Adding a custom method to the PostsController:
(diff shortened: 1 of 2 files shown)


parndt:
I don't agree. I think we should have app/decorators instead

kunalchaudhari:
I just followed convention.  For example model lives under app/models so is it's decorator. I think this will be easy to follow.

parndt:
My reasoning:
- Models live under app/models 
- Controllers live under app/controllers
- Views live under app/views
- Decorators live under app/decorators
- Presenters live under app/presenters

@radar I think Spree uses the conventions suggested by Kunal -- was this always the case?

Refinery has been using app/decorators for a while.. I guess we didn't get the memo to synchronise.

radar:
With Spree we have decorators inside their relative directories. I can see where @parndt is coming from and agree with it, but only if the structure is like this:

```
app/decorators
  |-> models
        |-> post_decorator.rb
  |-> controllers
        |-> posts_controller_decorator.rb
```

parndt:
Yes that is exactly how we do it in Refinery.

parndt:
We can get around this entire debate if we load app/*_/__decorator.rb but I'm not sure about that.

radar:
That's what @kunalchaudhari's initial patch does, but I would rather all the decorators not be mixed in with _real_ controllers/models/whatever.

parndt:
Right you are -- that is what it does.. I'm mainly referring to the guide that comes with the patch.

radar:
The guide should be patched to contain correct namespacing of the overrides as well, i.e.

```
app/decorators/controllers/forem/posts_controller_decorator.rb
```

Rather than:

```
app/decorators/posts_controller_decorator.rb
```

parndt:
100% agree.

radar:
:metal:

kunalchaudhari:
@radar missed the namespace in readme. actually had followed that in my project. Sorry for that. so we are going with spree like decorator and all decorators are namespaced?

radar:
That's right. Exactly as I've explained above.

parndt:
Correct so decorators live under app/decorators/

README.md
```diff
@@ -91,6 +91,32 @@ If you want to customise Forem, you can copy over the views using the (Devise-in
...
 You will then be able to edit the forem views inside the `app/views/forem` of your application. These views will take precedence over those in the engine.
 
+## Extending Classes
+
+All of Forem’s business logic (models, controllers, helpers, etc) can easily be extended / overridden to meet your exact requirements using standard Ruby idioms.
+
+Standard practice for including such changes in your application or extension is to create a file within the relevant app/models or app/controllers directory with the original class name with _decorator appended.
+
+### Adding a custom method to the Post model:
+    # app/models/forem/post_decorator.rb
```
parndt: Please make this `app/decorators/models/forem/post_decorator.rb`

README.md
```diff
@@ -91,6 +91,32 @@ If you want to customise Forem, you can copy over the views using the (Devise-in
...
+
+    Forem::Post.class_eval do
+      def some_method
+        ...
+      end
+    end
+
+### Adding a custom method to the PostsController:
+    # app/controllers/forem/posts_controller_decorator.rb
```
parndt: Please make this `app/decorators/controllers/forem/posts_controller_decorator.rb`

parndt:
Thank you very much -- I've merged this into 646c3e0 :-)

lib/forem/engine.rb
```diff
@@ -9,8 +9,13 @@ def root
       end
     end
 
-    # Fix for #88
-    config.to_prepare do
+    def self.activate
+
+      Dir.glob(File.join(File.dirname(__FILE__), "../app/**/*_decorator*.rb")) do |c|
+        Rails.configuration.cache_classes ? require(c) : load(c)
```
neerajsingh0101, line 15: Just curious . Any particular reason why this `require` vs `load` is being handled manually. Is Rails not equipped to unload these constants ?

radar, line 15: Legacy from Spree. I think you're right. We could probably get away with using `require_dependency` here?

atd:
FYI: I packaged this functionality as a gem, so it can be used by other engines. I also mentioned you as authors.

https://github.com/atd/rails_engine_decorators

PR #264, opened 2012-07-19
Use friendly_id gem to generate human friendly urls for categories, forums and topics. Fixes #262

### LINKED ISSUE #262: friendly_id for nice slugged urls
Would there be interest in including friendly_id to make all the url human (and search engine) friendlier? I can implement this feature and create a pull request if maintainers are interested. I'm planning to include this in my own project and would prefer this was part of forem. 

Any requests/preferences for implementation? I would like to add the friendly_id dependency to forem Gemfile and add the slug column to forums, categories and topics tables. Is that a good idea? Alternatively, I could create a new gem forem-friendly_ids that monkeypatches these models and generates migrations if that is the road you prefer. I would prefer to include it in core forem to make it simpler and this is a feature I imagine pretty much everyone would like.

juozasg:
Fixes https://github.com/radar/forem/issues/262

### DIFF
--- a/app/models/forem/forum.rb
+++ b/app/models/forem/forum.rb
@@ -1,6 +1,11 @@
+require 'friendly_id'
+
 module Forem
   class Forum < ActiveRecord::Base
     include Forem::Concerns::Viewable
+
+    extend FriendlyId
+    friendly_id :title, :use => :slugged
 
     belongs_to :category
 
--- a/db/migrate/20111103210835_create_forem_categories.rb
+++ b/db/migrate/20111103210835_create_forem_categories.rb
@@ -4,7 +4,5 @@ class CreateForemCategories < ActiveRecord::Migration
       t.string :name, :null => false
       t.timestamps
     end
-    Forem::Category.reset_column_information
-    Forem::Category.create(:name => 'General')
   end
 end
--- a/db/migrate/20120718073130_add_friendly_id_slugs.rb
+++ b/db/migrate/20120718073130_add_friendly_id_slugs.rb
@@ -1,0 +1,18 @@
+class AddFriendlyIdSlugs < ActiveRecord::Migration
+  def change
+    add_column :forem_forums, :slug, :string
+    add_index :forem_forums, :slug, :unique => true
+    Forem::Forum.reset_column_information
+    Forem::Forum.find_each {|t| t.save! }
+
+    add_column :forem_categories, :slug, :string
+    add_index :forem_categories, :slug, :unique => true
+    Forem::Category.reset_column_information
+    Forem::Category.find_each {|t| t.save! }
+
+    add_column :forem_topics, :slug, :string
+    add_index :forem_topics, :slug, :unique => true
+    Forem::Topic.reset_column_information
+    Forem::Topic.find_each {|t| t.save! }
+  end
+end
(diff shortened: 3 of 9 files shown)


app/models/forem/forum.rb
```diff
@@ -13,6 +16,7 @@ class Forum < ActiveRecord::Base
 
     attr_accessible :category_id, :title, :description, :moderator_ids
 
+
```
radar: Extra space here.

db/migrate/20111103210835_create_forem_categories.rb
```diff
@@ -4,7 +4,5 @@ def change
       t.string :name, :null => false
       t.timestamps
     end
-    Forem::Category.reset_column_information
-    Forem::Category.create(:name => 'General')
```
radar, line 8: Why?

db/migrate/20120718073130_add_friendly_id_slugs.rb
```diff
@@ -0,0 +1,18 @@
...
+
+    add_column :forem_categories, :slug, :string
+    add_index :forem_categories, :slug, :unique => true
+    Forem::Category.reset_column_information
+    Forem::Category.all.each {|t| t.save! }
+
+    add_column :forem_topics, :slug, :string
+    add_index :forem_topics, :slug, :unique => true
+    Forem::Topic.reset_column_information
+    Forem::Topic.all.each {|t| t.save! }
```
parndt: You can use find_each here instead of all.each

lib/forem/engine.rb
```diff
@@ -36,6 +36,7 @@ def self.activate
 end
 
 require 'simple_form'
+require 'friendly_id'
```
parndt: Can you please move this require to each model that uses it instead?

db/migrate/20111103210835_create_forem_categories.rb
juozasg, line 8: Moved this to seeds.rb for two reasons. First, including FriendlyId in Forem::Category will add a callback before_save which will look for slug column to update with correct slug. But at this point forem_category table does not have the slug column yet (that comes with the newest migration) so it's impossible to do things on Forem::Category (or Forem::Forum or Forem::Topic) before all migrations have been run.

Second reason is that this belongs to seed.rb anyway since it's seed data.

db/migrate/20120718073130_add_friendly_id_slugs.rb
juozasg: Certainly

lib/forem/engine.rb
juozasg: will do

juozasg:
Updated with changes except moving creating "General" Category to seeds.rb

db/migrate/20111103210835_create_forem_categories.rb
parndt, line 8: I'm :+1: on this change. No records should be added to the database in a migration.

lib/forem/engine.rb
parndt: Thanks a lot :-)

parndt:
Hey any chance you can implement the changes please? Thanks!

Oh I see, you did! They were just buried :-) thanks!

PR #268, opened 2012-07-24
Locales for PT, word added for EN and made the word By in a view use a Locale.

victormartins:


victormartins:
I don't understand why this is failing :/

radar:
It was failing beforehand. Sorry about that.

PR #273, opened 2012-07-30
fix #272 - fix undefined method user for nil:NilClass error when visiting forum which has pending review

### LINKED ISSUE #272: undefined method `user' for nil:NilClass when visiting forum after creation new topic(unapproved)
undefined method `user' for nil:NilClass when visiting forum after creation new topic(unapproved)

Follow this steps to produce
- Sign up new user or (login with existing user but forem_state must be equal to pending_review for that user)
- Visit any forum
- Create new topic
- Visit that forum again

kunalchaudhari:
fix undefined method user for nil:NilClass error when visiting forum which has pending review.

### DIFF
--- a/app/helpers/forem/topics_helper.rb
+++ b/app/helpers/forem/topics_helper.rb
@@ -23,6 +23,8 @@ module Forem
       posts = topic.posts.by_created_at.scoped
       if forem_admin_or_moderator?(topic.forum)
         posts
+      elsif topic.user == current_user
+        posts.visible.approved_or_pending_review_for(topic.user)
       else
         posts.approved
       end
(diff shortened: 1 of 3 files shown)


radar:
Could you write a regression test for this issue please?

avitus:
We're also getting this error. Appears be related to posts that haven't been moderated (?)

kunalchaudhari:
@avitus yeah. that happens because `relevant_posts(topic)` returns `[]` if topic has not been moderated yet. but if user who has started this post is logged in than topic should be displayed. And for other it won't show up as topic was not approved.

avitus:
Ok. It is keeping us motivated to get posts moderated quickly.

kunalchaudhari:
@avitus you can try out patch in my above commit.

radar:
Thank you @kunalchaudhari for reporting the issue and fixing it :)

Issue #274, opened 2012-08-01
How to access Forem routes outside of engine


murtazajafferji:
How do I use Forem url helpers outside of the engine in my main app? I get this error:

<code>undefined local variable or method `forums_path' for #<#Class:0x007fa1890db268:0x007fa1890b6260>
Extracted source (around line #2):
2:   <%= link_to t('forem.forum.forums'), forums_path %> &raquo; <%= link_to forum.category, forum.category %> &raquo; <%= link_to forum.title, forum %>
</code>

I have tried changing it to forums.forums_path, but that didn't work.

parndt:
Try forem.forums_path

murtazajafferji:
The reason why I thought forums.forums_path would work is because of this line in routes.rb:
<code>mount Forem::Engine, :at => "/forums"</code>

Anyways, I got this error when trying forem.forums_path:
<code>undefined method `forums_path' for #<#Class:0x007fa1890db268:0x007fa187a7e830>
Extracted source (around line #2):
2:   <%= link_to t('forem.forum.forums'), forem.forums_path %> &raquo; <%= link_to forum.category, forum.category %> &raquo; <%= link_to forum.title, forum %></code>

radar:
You should be using `forem.forums_path`. The engine's name is forem.

See the [Engines Guide](http://guides.rubyonrails.org/engines.html)

murtazajafferji:
In my last comment, I provided the error that occurs when I use <code>forem.forums_path</code>. Is there something else I should be doing?

murtazajafferji:
Do you think it has something to do with the fact that I am calling the url helpers inside of the Spree engine?

radar:
Calling the helpers inside Spree will be fine as long as you use the prefix. If you don't use the prefix then you will be having this problem. 

If you can duplicate the error in a brand new Rails app I can take a look at it once I am out of this traffic. 

## 

Ryan Bigg

murtazajafferji:
ok great. I just created a new app with Spree and Forem and I reproduced the issue:
https://github.com/murtazajafferji/forem_url_helpers_demo

radar:
Where in this application are you seeing the error?

murtazajafferji:
home#test. I set it up as the root.

radar:
The application is broken because `sign_in_path` is not defined.

murtazajafferji:
I added you as a collaborator. Do you mind pushing your fix up?

[...]

PR #279, opened 2012-08-03
add forem proxy in form builder missed in change of #274

kunalchaudhari:
all routes / form builders were prefixed with forem routes proxy in fix of #274. so one form builder is _post.html.erb file missed that change. updated that one.


radar:
Thanks! 

## 

Ryan Bigg

PR #286, opened 2012-08-15
[Javascript] Fixed autocompletion remote request failure for urls with leading slashes

gabriel-dehan:
Fixed autocompletion when Forem.base_path started with leading slashes which made the remote call not te contain host's url, leading to request failures.

Main reason was because remote calls failed when working with '/' as a base path.


parndt:
Fixed by merging your patch into 707e9519419b3c405c849e7c16f406acb1dd659b - thanks!

PR #294, opened 2012-08-30
CanCan's :read action is an alias to :index and :show actions

nashby:


radar:
I like the cut of your jib!

PR #296, opened 2012-09-01
don't use explicit `self` where it's not needed

nashby:


parndt:
What is wrong with being explicit?

nashby:
Actually, nothing. But it's like a common practice to avoid self where not required, e.g this style guide https://github.com/bbatsov/ruby-style-guide

radar:
I don't mind this change. I'll merge.

PR #297, opened 2012-09-04
Models cleanup

nashby:
- approved? method is granted by workflow gem
- it's possible to pass array to validates method
- remove obsolete code in approved scope

### DIFF
--- a/app/models/forem/post.rb
+++ b/app/models/forem/post.rb
@@ -92,10 +92,6 @@ module Forem
       user == other_user || other_user.forem_admin?
     end
 
-    def approved?
-      state == 'approved'
-    end
-
     protected
 
     def subscribe_replier
--- a/app/models/forem/post.rb
+++ b/app/models/forem/post.rb
@@ -39,10 +39,7 @@ module Forem
 
     class << self
       def approved
-        includes(:topic).
-        select("#{quoted_table_name}.*").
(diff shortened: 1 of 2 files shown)


radar:
Thanks for doing this!

nashby:
@radar no problem! :) 
BTW, what do you think about using https://github.com/ernie/squeel/ instead of using Arel directly? I can work on it if it sound OK to you.

radar:
What benefits would that provide?

nashby:
Much more cleaner syntax.

radar:
Do you have some examples of this?

nashby:
``` ruby
where("forem_topics.state = ? OR " +
      "(forem_topics.state = ? AND forem_topics.user_id = ?)",
       'approved', 'pending_review', user.id)
```

becomes 

``` ruby
where { (forem_topics.state = 'approved') | ((forem_topics.state = 'pending_review' ) & (forem_topics.user_id = my { user.id })) }
```

or something like this.

radar:
@parndt wdyt?

parndt:
I am in favour of not introducing more dependencies if it is just
syntactic sugar.  If we need to do something that squeel can do and
arel can't then it is a different story but I don't think this is the
case.

radar:
Ok, I can see that point in the argument. Makes sense to not include another dependency.

nashby:
Ok, I see.

PR #300, opened 2012-09-05
don't use arel were it's not needed

nashby:


knewter:
On the one hand - I think this looks nicer and is easier to read.  I think it's likely worth merging, but I know we have some uses of Arel idioms in here where it is actually needed to get performant queries (this likely not being one because omg)

+1

PR #315, opened 2012-10-03
Add an initializer to precompile a javascript asset...

knewter:
This was failing in prroduction if you tried to visit the members page.  The members page still acts funny, but this makes it at least visitable :)


knewter:
I'm merging this since no one else gave it love.  I'm using it in production, it works :)

PR #339, opened 2012-11-26
Add default_scope to order forums alphabetically

sfcgeorge:
Before this commit, forums were sorted by latest viewed at the bottom,
which caused much user confusion.

### DIFF
--- a/app/models/forem/forum.rb
+++ b/app/models/forem/forum.rb
@@ -20,6 +20,8 @@ module Forem
 
     alias_attribute :name, :title
 
+    default_scope order('title ASC')
+
     def last_post_for(forem_user)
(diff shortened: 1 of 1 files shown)


sfcgeorge:
It's only a tiny change, but I checked and tests are still passing. I haven't written a new test for this change as I'm not sure if it's needed.

sfcgeorge:
Travis CI emailed me to say I broke it. I'm not sure how one line of code broke it but apparently I managed it. I can't find where the error is in Travis, something on "DB=postgres  Ruby=1.8.7". Noob apology?

radar:
That was just a travis build artifact, nothing to be too concerned about.

I added a test for this patch and merged it with the above commit. Thank you!

Issue #253, opened 2012-06-26
uninitialized constant Forum 


danielres:
Hello

Just installed Forem on Rails 3.2.6.

I can see the forums list, I can enter the administration, manage forums and categories,...

But when I try to show a forum, I get an error:

``` ruby
NameError in Forem::ForumsController#show
uninitialized constant Forum
```

with:

``` ruby
Parameters:
{"id"=>"1"}
```

I tried with a blank app, same problem.

Any idea ?
Thanks

hazg:
cancan updated... 
gem 'cancan', '1.6.7' # can help you

agarie:
Using cancan 1.6.7 worked for me. Why is that? Could you explain? Thanks :)

mmcc:
+1 to both the issue and the solution.

jasonlfunk:
@agarie He meant that cancan updated to version 1.6.8 which broke Forem. If you force it to use 1.6.7, it works.

agarie:
@jasonlfunk Yes, it worked for me. Just wanted to know why it broke Forem. Thanks anyway!

parndt:
Often gems are released with backward incompatible changes in patch releases. We call these regressions and they're usually fixed once we let the author know of the problem who in this case is @ryanb :-)

Unless, of course, a bug was fixed in CanCan that we were relying on?

radar:
Can someone please provide a stacktrace for this issue? I'd be interested to know why this was broken.

radar:
Forem will now enforce CanCan 1.6.7 because 1.6.8 is broken. Will revise when necessary.

alexnesbitt:
changing forums_controller.rb from
-load_and_authorize_resource  :only => :show
+load_and_authorize_resource :class => 'Forem::Forum', :only => :show

seems to fix this for cancan 1.6.8

I think this was the change from 1.6.7
https://github.com/ryanb/cancan/commit/d5baed6281b4628071f8737090eaaec5168eacbf

parndt:
Seems like the change should be reverted as this makes CanCan harder to use?

radar:
Namespaced controllers may not mean namespaced resources, such as in the case of `Admin::ForumsController`. I don't mind this CanCan change. 

If we need to upgrade to a later version of CanCan we can make the change as @alexnesbitt suggests.

PR #353, opened 2012-12-13
Update CanCan to 1.6.8

vjt:
Hi,

we needed to update CanCan to 1.6.8, due to a dependency of the app in which we're integrating forem in, so we implemented and tested the solution proposed in issue #253. Tests pass on Rails 3.2.9.

It is OK for you to merge?

Thanks,

~Marcello, cc: @lleirborras

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -8,7 +8,7 @@ PATH
   remote: .
   specs:
     forem (1.0.0.beta1)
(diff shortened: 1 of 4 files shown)


radar:
Looks simple and the tests pass. Thank you!

vjt:
Great - thanks :smile:

PR #367, opened 2013-01-15
Added support for decorators and moved faux 'extensions' to decorators instead

parndt:


lib/forem/engine.rb
```diff
@@ -10,15 +10,7 @@ def root
     end
 
     config.to_prepare do
-      Dir.glob(Rails.root + "app/decorators/**/*_decorator*.rb").each do |c|
-        require_dependency(c)
-      end
-
-      require_dependency 'forem/user_class_extensions'
```
knewter, line 17: What was this doing and how did adding `decorators` allow us to remove it?  Or were those orthogonal?

parndt, line 17: I moved this -- note my rename above:

```
lib/forem/user_class_extensions.rb → app/decorators/lib/forem/user_class_decorator.rb
```

knewter, line 17: I totally skipped over that line, gotcha

Issue #301, opened 2012-09-06
how do I add members to groups?


sheam:
when I go to admin/groups/1 the add button is disabled. If I hit enter, it seems to accept my input anyway (bug?). 

Using an email, then hitting enter, it seems to be adding the user to the group, but I then get a missing template exception:
Missing template forem/admin/members/create, forem/application/create, application/create

There doesn't seem to be any documentation around the groups feature...

giedriusr:
:+1: I am facing with exactly the same issue and didn't get any feedback so far :/

bodrovis:
Seems to be fixed: https://github.com/radar/forem/pull/373

PR #373, opened 2013-01-30
FIx for Issue #301

0xCCD:


parndt:
@radar @knewter @robyurkowski can you forsee any issues with this patch?

radar:
No, I can't foresee any issues with this patch.

I think that we should go the way that Spree's solving this problem: defining a top-level `Spree` JavaScript variable and then [defining routes underneath that](https://github.com/spree/spree/blob/split_core/backend/app/views/spree/admin/shared/_routes.html.erb). That way, we wouldn't need this ugly Journey hack in place.

parndt:
Plus then we'd have support for any sub URI cleanly and with only exposing specific routes.

knewter:
+1

radar:
I've got a branch called members-select2 which is implementing this. It currently doesn't allow you to actually _add_ a member, which is problematic. To be looked at later.

knewter:
@radar @parndt So is the strategy to close this PR and await the top level js variable?  Is there an issue open for doing that?

radar:
@knewter: Yes, a top-level variable would be best. I don't think there's an open issue for fixing that other than #301.

daniely:
the link to defining routes by @radar is gone. i believe this is the [correct link](https://github.com/spree/spree/blob/master/backend/app/views/spree/admin/shared/_routes.html.erb)

bodrovis:
I am sorry, but was that ever fixed? I got the following error when visiting "Add a new member page": `TypeError: $(...).select2 is not a function` (I have created a brand new application with the latest Forem). Adding `gem "select2-rails"` directly to my Gemfile, `//= require select2` to application.js and `@import 'select2.css.scss';` to application.css.scss actually fixes this and everything works fine.

bodrovis:
Oh, I just added `//= require forem` to application.js and `@import 'forem/base.css';` to application.css.scss. This should be mentioned somewhere in the docs probably?

radar:
@bodrovis Yes, it should definitely be mentioned in the docs. Please submit a patch to fix that.

bodrovis:
@radar Done: https://github.com/radar/forem/pull/502

PR #375, opened 2013-02-06
Problem with Subscription#send_notification when user is removed

radar:
If the user object associated via subscription_id on the Subscription model is removed the send_notification method throws exceptions as it tries to pull the .id off a nil object.


parndt:
Thanks!

PR #380, opened 2013-02-13
Remove a negation in the italian translation

fracek:
The old translation was saying "Your reply has not been posted."

I also just noticed there are some inconsistencies in the italian translation, I will look into it. Does `en.yml` contain all the strings in Forem right?

### DIFF
--- a/config/locales/it.yml
+++ b/config/locales/it.yml
@@ -82,7 +82,7 @@ it:
     post:
       buttons:
         reply: Invia la risposta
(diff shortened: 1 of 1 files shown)


parndt:
@fracek that's correct, `en.yml` contains all the strings. It's the only one that all the maintainers speak unfortunately.

parndt:
Thanks for this solution.
Your code has not been merged! 

Wait.. I think I meant "Your code has been merged" ;-) hehe..

PR #383, opened 2013-02-23
Chinese translation (Traditional and Simplified)

ascendbruce:
Add Traditional Chinese (zh-TW) translation
Refine Simplified Chinese (zh-CN) translation. (some keys are missing)


parndt:
Thanks!

PR #385, opened 2013-02-24
Redo italian translation

fracek:
This is a pull request you can't refuse.
Now just go outside and enjoy yourself, and forget about this nonsense.


fracek:
On a more serious note, I stick with the phpBB translation even where I didn't agree with.

parndt:
Grazie!

PR #406, opened 2013-04-02
Refine i18n

ascendbruce:
Changes:
- Add missing translation lookup
- Add missing keys in en locale file
- Remove unused translation keys (which were added into en file in localeapp.com)

But there are some more problems not fix yet. eg. some translation scope are inconsistent or duplicated:
- `views/forem/admin/forums/new.html.erb` is using `t("forem.forum.new")`. but `views/forem/admin/groups/new.html.erb` is using `t("forem.admin.group.new")`.
- `forem.admin.forum.new` and `forem.forum.new` are duplicate. Only `forem.forum.new` is in use.

If there is a plan to fix it?

### DIFF
--- a/app/views/forem/admin/categories/new.html.erb
+++ b/app/views/forem/admin/categories/new.html.erb
@@ -1,3 +1,3 @@
-<h2><%= t("forem.admin.category.new") %></h2>
+<h2><%= t(".create_category") %></h2>
 
 <%= render "form" %>
--- a/app/views/forem/admin/categories/new.html.erb
+++ b/app/views/forem/admin/categories/new.html.erb
@@ -1,3 +1,3 @@
-<h2><%= t("forem.category.new") %></h2>
+<h2><%= t("forem.admin.category.new") %></h2>
 
 <%= render "form" %>
--- a/app/views/forem/admin/categories/edit.html.erb
+++ b/app/views/forem/admin/categories/edit.html.erb
@@ -1,3 +1,3 @@
-<h2><%= t("forem.admin.category.edit", :title => @category.name, :default => "Editing the %{title} Forum Category") %></h2>
+<h2><%= t(".edit_category", :title => @category.name, :default => "Editing the %{title} Forum Category") %></h2>
 
 <%= render "form" %>
--- a/app/views/forem/admin/categories/edit.html.erb
+++ b/app/views/forem/admin/categories/edit.html.erb
@@ -1,3 +1,3 @@
-<h2><%= t("forem.category.edit", :title => @category.name) %></h2>
+<h2><%= t("forem.admin.category.edit", :title => @category.name, :default => "Editing the %{title} Forum Category") %></h2>
 
 <%= render "form" %>
--- a/app/views/forem/admin/categories/index.html.erb
+++ b/app/views/forem/admin/categories/index.html.erb
@@ -5,17 +5,17 @@
 <table id='forums' class='forums subdued' cellspacing='0'>
   <thead>
     <tr>
-      <th><%= t('edit', :scope => 'forem.admin.categories') %></th>
-      <th><%= t('delete', :scope => 'forem.admin.categories') %></th>
-      <th><%= t('category', :scope => 'forem.admin.categories') %></th>
+      <th><%= t('.edit') %></th>
+      <th><%= t('.delete') %></th>
+      <th><%= t('.category') %></th>
 
     </tr>
   </thead>
   <tbody>
     <% @category.each do |category| %>
       <tr class="forum <%= cycle("odd", "even") %>">
-        <td><%= link_to t('edit', :scope => 'forem.admin.categories'), forem.edit_admin_category_path(category) %></td>
-        <td><%= link_to t('delete', :scope => 'forem.admin.categories'), forem.admin_category_path(category), :method => :delete, :confirm => t("delete_confirm", :scope => "forem.admin.categories") %></td>
+        <td><%= link_to t('.edit'), forem.edit_admin_category_path(category) %></td>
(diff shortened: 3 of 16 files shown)


app/views/forem/admin/categories/new.html.erb
```diff
@@ -1,3 +1,3 @@
-<h2><%= t("forem.category.new") %></h2>
+<h2><%= t("forem.admin.category.new") %></h2>
```
parndt: For things like this we should just be able to specify with a leading period:

``` erb
<h2><%= t('.new') %></h2>
```

I18n should figure out the rest of the namespace based on the view path `forem/admin/categories/ -> forem.admin.categories`

I'm not sure why it's under category in the first place and not categories. This breaks conventions. If I did it, sorry.

ascendbruce: I will fix it later.

I plan to change only `en.yml` file. And after updating translation keys in locateapp.com, than update Chinese translation. Is that ok?

parndt: OK, I'm keen to have this fixed before we merge this though as people do use the master branch as a reliable source (has worked so far!)

ascendbruce: I found that most of lookups for title didn't use lazy lookup. eg. [admin/admin/topics/edit](https://github.com/radar/forem/blob/master/app/views/forem/admin/topics/edit.html.erb), [admin/forums/new](https://github.com/radar/forem/blob/master/app/views/forem/admin/forums/new.html.erb) and [topics/new](https://github.com/radar/forem/blob/master/app/views/forem/topics/new.html.erb).

To fit the same style in those files above. Maybe `forem.category.new` is the right answer. If so, we should add translation key for it.

Is there a conventions guidelines for this project?

radar: There's no conventions guideline for the project. The code is the guideline.

wrt to I18n, I think matching the translations up to their dot-prefix equivalents is the right thing to do. i.e. `<%= t(".forum") %>` should look up the correct translation relative to that view.

ascendbruce: I'd like to change the translation key to: `forem.admin.categories.new.create_category`

than in `app/views/forem/admin/categories/new.html.erb`:

``` erb
<h2><%= t('.create_category') %></h2>
<!-- translation is: "Creating a new Forum Category" -->
```

And update all similar cases. (`topics/new`, `admin/topics/edit`, `admin/forums/new`, etc...)

parndt: Makes sense to me.

ascendbruce:
In commit [fcdf4c9](https://github.com/ascendbruce/forem/commit/fcdf4c92057743582af96fc3a12a14fe56c2563a), I updated most similar cases. The rests are a little more complicated. I'd like to leave it for a big refactoring.

knewter:
@radar @parndt I've been out of the loop, but I've seen some movement in the translations area lately.  Anything that should keep me from clicking the big green 'merge' button? :+1:

parndt:
@radar so when we merge this what happens to the localeapp stuff? How is that kept syncronised currently?

ascendbruce:
in [localeapp's document](https://github.com/Locale/localeapp/wiki/Configure-the-Locale-Ruby-Gem): (with gem configured)

> By default, your Ruby on Rails application will only send newly created translation keys when running in the development environment. When a page is refreshed locally, any unpopulated keys are sent to the app's corresponding localeapp.com project.

or manually way: [Import & Export Translations Files](https://github.com/Locale/localeapp/wiki/Import-&-Export-Translations-Files)

Note that in commit [fcdf4c9](https://github.com/ascendbruce/forem/commit/fcdf4c92057743582af96fc3a12a14fe56c2563a), I only updated `en` translation file. I am not familiar with localeapp.com workflow. But I guess it will be good to sync it before or short after merged.

ps. I prefer manually way, ~~to avoid receiving keys from everywhere. (and I don't know what will happen if I18n::MissingTranslation is not raised from forem code.)~~ Seems that [it requires API_KEY to connect to localeapp project](https://github.com/Locale/localeapp/wiki/Install-the-Locale-Ruby-Gem). So it won't happen.

radar:
@ascendbruce I'm fine with having the dependency on `localeapp`, just as long as we document somewhere that people need to generate an API key locally before they can run Forem.

Reckon you could update this patch to include that?

radar:
On second thought, maybe not a good idea. Just read up some more on their documentation. Maybe we should have it set `polling_environments` and `sending_environments` to `[]` by default, and only have it set if an env variable like `TRANSLATING` is set.

ascendbruce:
As i understand, the API key is only available to localeapp project owner and developers. So most contributors can't generate it.

I created an app to try out localeapp.com. But I'm still facing some problems such as It doesn't send any unpopulated keys, even when `config.sending_environments = [:development]` is set explicitly

glebm:
I've recently released a gem with tasks to find missing and unused translations, should be a perfect fit for forem: https://github.com/glebm/i18n-tasks

parndt:
@radar @knewter I have no idea why we didn't merge this one

radar:
I don't know either. Maybe because it doesn't apply cleanly any more.

Hey @ascendbruce: could you please tidy up this PR so that it applies cleanly?

ascendbruce:
Yes but it may take days :sweat_smile:

radar:
That's alright. We are patient :)

radar:
@ascendbruce Any updates on this?

ascendbruce:
Hi! @radar 

Sorry I've been busy these days. Still working on it but not yet complete :disappointed:

radar:
@ascendbruce No worries. Just let me know when you're done :)

PR #407, opened 2013-04-09
Partially completed Japanese localizations

BanzaiMan:
Some parts are missing, since they require either the context in which they appear or good translations.


radar:
Thanks for your work on this Hiro :)

PR #409, opened 2013-04-13
Modified gem to allow page number rendering in reply_to links

abrambailey:
Added @post_positions to topics show action:

``` ruby
def show
  if find_topic
    register_view
    @posts = @topic.posts
    unless forem_admin_or_moderator?(@forum)
      @posts = @posts.approved_or_pending_review_for(forem_user)
    end
    @posts = @posts.page(params[:page]).per(Forem.per_page)
    #Provides ids of posts and their positions
    post_positions = @topic.posts.flatten.collect{|p| p.id}.each_with_index.to_a
    @post_positions = post_positions.map{|a| {:id=>a[0], :position=>a[1]+1}}
  end
end
```

Modified posts/_post.html.erb view:

``` erb
<% if post.reply_to %>
  <span class='in_reply_to'>
    <% position = @post_positions.select{|p| p[:id].equal? post.reply_to.id}.first[:position] %>
    <% page = Integer(Float(position-1)/Float(Forem.per_page)+1) %>
    <%= link_to "#{t("forem.post.in_reply_to")} #{post.reply_to.user}", "?page=#{page}#post-#{post.reply_to.id}" %>
  </span>
<% end %>
```

Modified topics/show.html.erb to pass data to partial:

``` erb
<div id='posts'>
  <%= render :partial => "forem/posts/post", :collection => @posts, :locals => {:post_positions=>@post_positions} %>
</div>
```

### DIFF
--- a/app/views/forem/posts/_post.html.erb
+++ b/app/views/forem/posts/_post.html.erb
@@ -31,7 +31,9 @@
 
     <% if post.reply_to %>
       <span class='in_reply_to'>
-        <%= link_to "#{t("forem.post.in_reply_to")} #{post.reply_to.user}", "#post-#{post.reply_to.id}" %>
+        <% position = @post_positions.select{|p| p[:id].equal? post.reply_to.id}.first[:position] %>
+        <% page = Integer(Float(position-1)/Float(Forem.per_page)+1) %>
+        <%= link_to "#{t("forem.post.in_reply_to")} #{post.reply_to.user}", "?page=#{page}#post-#{post.reply_to.id}" %>
       </span>
     <% end %>
 
--- a/app/controllers/forem/topics_controller.rb
+++ b/app/controllers/forem/topics_controller.rb
@@ -79,6 +79,12 @@ module Forem
         redirect_to @forum and return
       end
     end
+    
+    def post_count(topic)
+      a=topic.posts.flatten.collect{|p| p.id}.each_with_index.to_a
+      a=a.map{|a| {:id=>a[0], :position=>a[1]+1}}
+      return a
+    end
 
     def register_view
       @topic.register_view_by(forem_user)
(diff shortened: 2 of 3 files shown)


radar:
Not sure how I feel about all these calculations going on. I would rather the calculation be done once, i.e. when the post is saved, rather than every time the post is viewed. After all, it's unlikely that the post's page number is going to change. Agree?

parndt:
Extremely unlikely. The only situation is where someone deleted a post you'd need to renumber each one.. or when you changed how many posts per page.

abrambailey:
I suppose you could also store the small numbering array in the Rails cache with topic_id in the name and delete and re-create the cache item given any of the above situations ... Rather than update the db records.. not sure if this is an alternative or not.

knewter:
So I would expect the page number to change because I'm used to 'per page' being configurable for the end user.  Unless it's actually _slow_ I'd be fine with it, if the actual calculations were pulled out into a helper (logic in views is scary)

app/views/forem/posts/_post.html.erb
```diff
@@ -31,7 +31,9 @@
 
     <% if post.reply_to %>
       <span class='in_reply_to'>
-        <%= link_to "#{t("forem.post.in_reply_to")} #{post.reply_to.user}", "#post-#{post.reply_to.id}" %>
+        <% position = @post_positions.select{|p| p[:id].equal? post.reply_to.id}.first[:position] %>
+        <% page = Integer(Float(position-1)/Float(Forem.per_page)+1) %>
```
parndt, line 35: This and the position calculation don't belong in the view. :)

parndt:
@knewter yeah, you're right.. the `per_page` should be able to be configurable which means that any pre-calculation of it is not appropriate. Is there any way that we can just provide a routing helper that does all of this?

radar:
Agreed on the points @parndt and @knewter make.

@funkdified: Could you please move the calculation code into a helper and add some specs around it too? That'd be great.

abrambailey:
Hi, just noticed this comment thread again... Sorry been slammed at work. I will have a look at this during the week.

PR #411, opened 2013-04-24
Add Vietnames translation

runlevel5:


radar:
Hey mate, thanks for these. The first line in that file should be the locale symbol, probably "vi" rather than "en"

runlevel5:
@radar re-submit new one now

Issue #146, opened 2012-02-14
Page titles


glebm:
The page titles are all the same, which Google (= The Web God) doesn't like.

We came up with a workaround like this: 

```
- title = ""
- if @topic
   - title += "#{@topic.subject} - #{@topic.forum.title} - #{@topic.forum.category.name} forums - "
- elsif @forum
  - title += "#{@forum.category.name} forums - #{@forum.title} - "
- elsif @category
  - title += "#{@category.name} forums - "
#{title}Site Name
```

It would be nice to have some way of customizing them without resorting to using the controller's instance variables though.

One suggestion would be to use I18n.

parndt:
A helper would be better than introducing this conditional logic in the view :) I'm not sure how that allows customisation though. @knewter any thoughts?

Besides Google, I'm pretty sure ordinary internet uses like titles to differ too.

knewter:
So yeah, the traditional way we've handled this is with a helper....so our application layout tends to have:

```
%title= render_title
```

Where there's a 

```
def render_title
  [@title, @site_name].compact.join(" | ")
end

def title=(title)
  @title = title
end
```

Then in our views or controllers, we'll just use `title=` to set the title.  Does that make sense?

glebm:
@knewter However, forem controllers do not set @title 

The idea I'm proposing is to have titles come from I18n interpolations. E.g., if forem does something like this (and exposes it somehow, e.g. forem.page_title):

```
I18n.t('forem.topic.title', topic: @topic.subject, forum: @topic.forum.title)
```

Then you gain sensible default titles and the ability to customize them.

Now, this is just the first idea, I am sure we can come with a better one.

PS

View vs helper is not the point here (that code was just an example)
Everyone has different techniques for this kind of thing. 
For example, We use 

```
%title= content_for?(:page_title) ? yield(:page_title) : site_name
```

And then in the view

```
- content_for :page_title, "My Account - #{site_name}"
```

knewter:
Right, so I'm at SpreeConf and completely not in-context  :)  I think using refinery for i18n page titles where they come from refinery, and locale files for pages that don't fall under that purview, makes plenty of sense...

parndt:
@knewter this is Forem not Refinery ;-)

knewter:
doesn't matter... :-\  I should've stopped when I said I wasn't in-context obviously :)

The proposal @glebm made is sensible.  Why don't we do it that way?

[...]

PR #413, opened 2013-04-28
Display page titles

fracek:
I implemented page titles as indicated in #146, every view calls an helper to set the title for that page.

### DIFF
--- a/app/helpers/forem/application_helper.rb
+++ b/app/helpers/forem/application_helper.rb
@@ -49,5 +49,9 @@ module Forem
         end
       end.html_safe if content.present?
     end
+
+    def set_forem_page_title(page_title)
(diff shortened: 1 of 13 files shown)


fracek:
Sorry I thought forem was ruby 1.9+.

Do you want me to change to use the hashrocket and make another pull request? To maintain a clean history.

radar:
No need for another pull request. Please just revert to using hashrocket syntax and force-push a fixed commit to your branch. That will update this pull request.

parndt:
Does this have to happen in the view template?

fracek:
I think it's where it makes more sense, if you think it's better doing it another way I will change it.

Spaceghost:
You could put it on the view object itself if you like.

PR #419, opened 2013-05-09
Split up the responsibilities of the PostsController

parndt:
a) I can actually understand it
b) I can override particular parts of it with decorators like the *_successful and *_failed hooks


knewter:
I wholeheartedly approve this change.  Thanks!

PR #422, opened 2013-05-19
Run the dummy app with localeapp when an API key is supplied.

parndt:
Is this remotely approaching correct?


radar:
What is this supposed to fix?

parndt:
The sync of localeapp

parndt:
@radar thoughts, comments?

radar:
Seems fine to me

parndt:
Localeapp is dead, long live Forem without Localeapp!

PR #423, opened 2013-05-19
Decoupled actions from their callbacks.

parndt:


parndt:
@knewter Ryan's probably busy and has inbox 5000 so it's up to you :)

radar:
A++ would merge again.

knewter:
Was in the air all day so I missed my chance

PR #424, opened 2013-05-20
Add :dependent => :destroy to topic subscriptions and views

yan-hoose:
If you use foreign key constraints in your database then it is not possible to delete a topic, since rows in other tables depend on the topic. 

This commit adds :dependent => :destroy to topic subscriptions and views which makes it possible.

### DIFF
--- a/spec/models/topic_spec.rb
+++ b/spec/models/topic_spec.rb
@@ -20,6 +20,24 @@ describe Forem::Topic do
     end
   end
 
+  context "deletion" do
+    it "deletes views, subscriptions and posts" do
+      FactoryGirl.create(:post, :topic => @topic)
+      FactoryGirl.create(:subscription, :topic => @topic)
+      FactoryGirl.create(:topic_view, :viewable => @topic)
+
+      @topic.posts.count.should be > 0
+      @topic.subscriptions.count.should be > 0
+      @topic.views.count.should be > 0
+
+      @topic.destroy
+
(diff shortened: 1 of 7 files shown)


spec/models/topic_spec.rb
```diff
@@ -20,6 +20,24 @@
...
+
+      @topic.posts.count.should be > 0
+      @topic.subscriptions.count.should be > 0
+      @topic.views.count.should be > 0
+
+      @topic.destroy
+
+      Forem::Post.where('topic_id = ?', @topic.id).first.should be_nil
+      Forem::Subscription.where('topic_id = ?', @topic.id).first.should be_nil
+      Forem::View.where('viewable_id = ?', @topic.id).first.should be_nil
```
radar: I think using the exists method here would be better than using where.

spec/models/topic_spec.rb
```diff
@@ -20,6 +20,24 @@
...
 
+  context "deletion" do
+    it "deletes views, subscriptions and posts" do
+      FactoryGirl.create(:post, :topic => @topic)
+      FactoryGirl.create(:subscription, :topic => @topic)
+      FactoryGirl.create(:topic_view, :viewable => @topic)
+
+      @topic.posts.count.should be > 0
+      @topic.subscriptions.count.should be > 0
+      @topic.views.count.should be > 0
```
parndt, line 31: Are these really necessary? We can see you creating the associations above. What are they for?

Issue #427, opened 2013-05-21
Deleted users and stranded posts/topics


avitus:
Looking through the documentation I can't seem to find anything about how to handle deleted users. Currently, forem throws an exception when trying to display the link and avatar for a user that no longer exists.

How are other users handling this? Is it best practice to delete all associated posts/topics. It seems as though removing entire topics created by deleted users is unnecessary but the fact that the Forem engine seems to think that should be the default is making us wonder.

parndt:
Interesting. Probably our users would have to be soft deleted through something like paranoia. What do you think, everyone else? @knewter @radar

knewter:
I would suggest soft deletion

radar:
Soft deletion also gets my vote.

PR #431, opened 2013-05-27
Implement 'Nil User' pattern to deal with deleted users

radar:
Potential fix for #427.

I think this is a good way to deal with deleted users. @knewter, @parndt, OK?

### DIFF
--- a/app/views/forem/posts/_post.html.erb
+++ b/app/views/forem/posts/_post.html.erb
@@ -3,7 +3,11 @@
 
   <div class='user'>
     <div class='username'>
-      <%= link_to_if Forem.user_profile_links, post.user, [main_app, post.user] %>
+      <% if post.user.is_a?(Forem::NilUser) %>
+        <%= t(:deleted) %>
+      <% else %>
+        <%= link_to_if Forem.user_profile_links, post.user, [main_app, post.user] %>
+      <% end %>
     </div>
     <div class='icon'><%= forem_avatar(post.user, :size => 60) %></div>
   </div>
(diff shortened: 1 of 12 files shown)


parndt:
In general seems good.

radar:
I've snuck in another change here which was required to fix some of the tests for the nil user.

It updates Forem to use Workflow 1.0.0 and switches some update_attribute calls to update_column calls so that the callbacks aren't triggered _all the damn time_. Workflow also provides the ability to define event methods which allows us to avoid using typical ActiveRecord callbacks for our state machine events.

spec/support/factories/topics.rb
```diff
@@ -5,10 +5,10 @@
     t.user {|u| u.association(:user) }
     t.posts_attributes { [:text => "This is a brand new post"] }
 
-    trait :approved do
-      state 'approved'
+    factory :approved_topic do
+      after_create do |t|
+        t.approve!
```
radar, line 10: This is necessary to trigger the `approve` method callbacks.

app/views/forem/posts/_post.html.erb
```diff
@@ -3,7 +3,11 @@
 
   <div class='user'>
     <div class='username'>
-      <%= link_to_if Forem.user_profile_links, post.user, [main_app, post.user] %>
+      <% if post.user.is_a?(Forem::NilUser) %>
+        <%= t(:deleted) %>
+      <% else %>
+        <%= link_to_if Forem.user_profile_links, post.user, [main_app, post.user] %>
+      <% end %>
```
parndt, line 10: Normally a NilUser would make the view less complicated ;-)

avitus:
Would be great if this could be merged.

parndt:
My only gripe was the view complication.

radar:
Added to master and rails4. If anyone has any ideas on how to make the view less complicated, patches are welcome :)

Issue #429, opened 2013-05-25
Moderation Queue Notifications?


createthis:
Hey,

Are Admins notified when a new message hits the moderation queue? I'm not seeing any emails coming from my new installation on this event.

Thanks,

Jesse

radar:
There are currently no email notifications for this. I agree that it would be a good idea to email admins regarding these events.

createthis:
Yeah, I've already had a few days slip by while I had messages in the queue unawares. I think this needs to happen. Seems like it would be a really easy feature to implement too. If you can point me to the most likely trigger point real quick, I can probably scratch my own itch and give you a pull request.

parndt:
We'd welcome a pull request with open arms. The logic in `skip_pending_review` could probably be extended to handle the notification case too.

https://github.com/radar/forem/blob/master/app/models/forem/post.rb#L113-L115

createthis:
Well, that was certainly a starting point, but after another hour of digging, I think what I was looking for was Forem::TopicsController#create using @topic.user.forem_moderate_posts? as a filter. I'll see what I can do.

createthis:
One problem I'm seeing is the lack of a clear way to iterate moderators. There is a has_many :moderators in the Forem::Forum model that may be intended to fill this need, but I'm not sure it is working or complete. Instead, I decided to build a list of administrators.

createthis:
Scratch that pull request. The post code is firing on update as well as create. I'll have to move it.

createthis:
Alright. I apparently don't have a clue how these pull requests work. I added a commit to fix the code on my end, and it seems to have automatically updated the pull request. Fancy. We should be good to go now.

PR #433, opened 2013-06-02
 Issue #429 - Moderation Queue Notifications?

createthis:
Issue #429 - Moderation Queue Notifications?

### DIFF
--- a/app/controllers/forem/posts_controller.rb
+++ b/app/controllers/forem/posts_controller.rb
@@ -21,6 +21,7 @@ module Forem
       @post.user = forem_user
 
       if @post.save
+        ModerationQueueMailer.new_post(@post).deliver if @post.topic.user.forem_moderate_posts?
         create_successful
       else
         create_failed
--- a/app/controllers/forem/topics_controller.rb
+++ b/app/controllers/forem/topics_controller.rb
@@ -23,6 +23,7 @@ module Forem
       @topic = @forum.topics.build(params[:topic], :as => :default)
       @topic.user = forem_user
       if @topic.save
+        ModerationQueueMailer.new_topic(@topic).deliver if @topic.user.forem_moderate_posts?
         create_successful
       else
         create_unsuccessful
--- a/app/mailers/forem/moderation_queue_mailer.rb
+++ b/app/mailers/forem/moderation_queue_mailer.rb
@@ -1,0 +1,20 @@
+module Forem
+  class ModerationQueueMailer < ActionMailer::Base
+    default :from => Forem.email_from_address
+
+    def new_topic(topic)
+      @topic = topic
+      
+      @admin_users = Forem.user_class.find_all_by_forem_admin(true)
+      mail(:to => @admin_users.map(&:email), :subject => I18n.t('forem.topic.moderation_needed'))
+    end
+
+    def new_post(post)
+      @post = post
+      
+      @admin_users = Forem.user_class.find_all_by_forem_admin(true)
+      mail(:to => @admin_users.map(&:email), :subject => I18n.t('forem.post.moderation_needed'))
+    end
+  end
+end
+
--- a/app/models/forem/post.rb
+++ b/app/models/forem/post.rb
@@ -38,6 +38,7 @@ module Forem
     after_save :approve_user,   :if => :approved?
     after_save :blacklist_user, :if => :spam?
     after_save :email_topic_subscribers, :if => Proc.new { |p| p.approved? && !p.notified? }
+    after_save :send_moderation_notification, :if => Proc.new { |p| p.topic.user.forem_moderate_posts? }
 
     class << self
       def approved
@@ -97,6 +98,10 @@ module Forem
       if topic && user
         topic.subscribe_user(user.id)
       end
+    end
+
+    def send_moderation_notification
+      ModerationQueueMailer.new_post(self).deliver if topic.user.forem_moderate_posts?
     end
 
     def email_topic_subscribers
--- a/app/models/forem/post.rb
+++ b/app/models/forem/post.rb
@@ -38,7 +38,6 @@ module Forem
     after_save :approve_user,   :if => :approved?
     after_save :blacklist_user, :if => :spam?
     after_save :email_topic_subscribers, :if => Proc.new { |p| p.approved? && !p.notified? }
-    after_save :send_moderation_notification, :if => Proc.new { |p| p.topic.user.forem_moderate_posts? }
 
     class << self
       def approved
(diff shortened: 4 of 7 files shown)


app/controllers/forem/topics_controller.rb
```diff
@@ -23,6 +23,7 @@ def create
       @topic = @forum.topics.build(params[:topic], :as => :default)
       @topic.user = forem_user
       if @topic.save
+        ModerationQueueMailer.new_topic(@topic).deliver if @topic.user.forem_moderate_posts?
```
parndt, line 26: This should go in `create_successful` so that people can easily override this part..

app/controllers/forem/posts_controller.rb
```diff
@@ -21,6 +21,7 @@ def create
       @post.user = forem_user
 
       if @post.save
+        ModerationQueueMailer.new_post(@post).deliver if @post.topic.user.forem_moderate_posts?
```
parndt, line 24: This should go in `create_successful` so that people can easily override this part..

parndt:
Can you wrap the call to the mailer and its predicate inside methods? 

e.g.

``` ruby
protected
def create_successful
  deliver_moderation_notification if deliver_moderation_notifications_for_user?(@post.topic.user)
  # what is currently in the create_successful method
end

def deliver_moderation_notification
  ModerationQueueMailer.new_post(@post).deliver
end

def deliver_moderation_notifications_for_user?(user)
  user.forem_moderate_posts?
end
```

parndt:
This has made me think further around moderation.. is the only moderator really the `@post.topic.user`? Does this user want to be emailed? Should we have a list of moderators?

parndt:
cc @radar @knewter

createthis:
Um, no, the moderator is defined by:

``` ruby
@admin_users = Forem.user_class.find_all_by_forem_admin(true)
```

I don't see the value in the other suggestions. If I need to override something, I fork the code and make my own changes. I don't see why we need to wrap everything in method calls. It just gets confusing.

parndt:
That's great that you're comfortable forking everything.. but not everyone wants to do this as it is far easier/safer/maintainable to just decorate the class and override one method than to have to modify unrelated functionality. The ultimate would be to extract the entire thing into a single responsibility class like `NotifiesModerators` or similar which took in the post and handled the entire process as this is a responsibility that [doesn't ultimately belong in the controller](http://robots.thoughtbot.com/post/50655960596/sandi-metz-rules-for-developers). 

You may have to take my word for it that making the software more maintainable makes our lives a lot easier.. chalk it up to lessons learned over 6+ years of open source programming if you like. :smile:

parndt:
So if we're simply deciding whether to moderate the current user's posts we really can just pass `forem_user` instead of `@post.topic.user` is that correct?

``` ruby
def create_successful
  deliver_moderation_notification if deliver_moderation_notifications_for_user?(forem_user)
  # what is currently in the create_successful method
end
```

parndt:
By the way I should mention that I love the feature (which is why I'm paying so much attention to it) I'm just trying to reduce future headaches for us. :heart:

createthis:
I haven't bought into the whole decorator class system yet. I admittedly use decorators from time to time, but I'm concerned they make things MORE difficult to maintain long term, not easier. They encourage "spaghetti code", which is something I thought we got rid of in the previous decade. For example, I just spent hours debugging something caused by a decorator _I wrote_. It took that long just to figure out where the problem was being introduced. I think they cause more problems than they solve. Use them cautiously.

13+ years as a professional software developer has taught me to only take things as far as the immediate requirements dictate. Over-engineering for future issues that may never exist is a trap. It's particularly a trap for intuitive dominant engineers, like you and me. Sensors rarely think so far ahead. Sometimes I think software should be a Sensor dominated field because of this. We intuitives need to take lessons from them.

Finally, I'm convinced this code belongs in the controller. It's fine there. I think it's a good idea to put the bulk of logic into a model, but notifications like this definitely belong in the controller. You can even think of them as an alternative view if it helps. An email view.

Look, you're clearly more invested in this project than I am. I just scratched my itch. You're welcome to run with it at this point. I've got other projects to work on. Good luck!

app/mailers/forem/moderation_queue_mailer.rb
```diff
@@ -0,0 +1,20 @@
+module Forem
+  class ModerationQueueMailer < ActionMailer::Base
+    default :from => Forem.email_from_address
+
+    def new_topic(topic)
+      @topic = topic
+      
+      @admin_users = Forem.user_class.find_all_by_forem_admin(true)
```
radar, line 8: This should additionally email all the moderators for the topic's forum: `topic.forum.moderators`. Both the moderators and the admins are able to approve posts. A topic's owner cannot approve.

app/mailers/forem/moderation_queue_mailer.rb
```diff
@@ -0,0 +1,20 @@
...
+      mail(:to => @admin_users.map(&:email), :subject => I18n.t('forem.topic.moderation_needed'))
+    end
+
+    def new_post(post)
+      @post = post
+      
+      @admin_users = Forem.user_class.find_all_by_forem_admin(true)
```
radar, line 15: I suggest moving this into its own method.

app/mailers/forem/moderation_queue_mailer.rb
createthis, line 8: I made a comment about this earlier:

> One problem I'm seeing is the lack of a clear way to iterate moderators. There is a has_many :moderators in the Forem::Forum model that may be intended to fill this need, but I'm not sure it is working or complete. Instead, I decided to build a list of administrators.

I couldn't find any documentation on the moderators functionality. There's the moderator group table... but I couldn't get it working. Do you have any insight to offer?

app/mailers/forem/moderation_queue_mailer.rb
createthis, line 15: How do you mean?

radar, line 15:
``` ruby
def admin_users
  Forem.user_class.find_all_by_forem_admin(true)
end
```

The reason being is that it's referenced in both the new_topic and new_post method. I'll do this when I merge it, one moment.

app/mailers/forem/moderation_queue_mailer.rb
radar, line 8: Use `>` for quotes, not backticks.

---

You can add users to groups in the admin backend. The groups can then be added to a forum as moderators. This functionality needs another look at, I think. It's a bit rough around the edges.

radar:
I've merged this code into the `moderation-queue` branch. There's still work to be done, so I will open a new PR with a checklist of what needs doing and then we can go from there.

radar:
New PR is #434. Thank you for all your work on this @createthis!

PR #434, opened 2013-06-03
Moderation queue

radar:
This is the work from #433, moved to the main repository since the [original owner gave up](https://github.com/radar/forem/pull/433#issuecomment-18812715).

---
- [x] Write tests for TopicsController moderation queue mailers
- [ ] Write tests for the mailer itself.
- [ ] Ensure users who are within a forum's `moderators` group get emailed.
- [ ] Include post text within moderation email, unlike Google Groups.

### DIFF
--- a/app/controllers/forem/posts_controller.rb
+++ b/app/controllers/forem/posts_controller.rb
@@ -21,6 +21,7 @@ module Forem
       @post.user = forem_user
 
       if @post.save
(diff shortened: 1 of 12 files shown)


parndt:
Improved!

parndt:
This is an obvious place where we can maybe DRY up the entire moderation email process, given that both controllers are doing remarkably similar things.  But we can do that later.

radar:
Looking good!

parndt:
Over to you again @radar

PR #439, opened 2013-06-15
Add configurable Forem#base_path

parndt:
For example, in the case where I mount under my engine instead of the main application
the previous code wasn't working and I had to use:

``` ruby
Forem.base_path = MyExtension::Engine.routes.named_routes[:forem].path
```

(which is now possible)


parndt:
OK so CI randomly fails since maybe b44bd73744413602759a817e48ca78fecce97b4b

parndt:
Merged it.

Issue #445, opened 2013-07-16
Should users be able to reply without having read permissions on forum?


daniely:
https://github.com/radar/forem/blob/master/app/models/forem/ability.rb#L44

How come there is no check to see if a user has read permissions on a forum before allowing them to reply to a topic? Is this not a use case to worry about?

For example, say a user at one point has access to a forum (and thus topics). Eventually they get banned for bad behavior. If they are clever enough I believe they'd be able to assemble an http POST request to reply to an existing topic, even though they are banned.

If this is something that needs to be fixed I'd be happy to write a test and submit a patch.

parndt:
Could this be achieved by overriding this method in your user's class? https://github.com/radar/forem/blob/master/lib/forem/default_permissions.rb#L32-L36

``` ruby
def can_reply_to_forem_topic?(topic)
  can_read_forem_forum?(topic.forum) && true # or some condition in place of true
end
```

daniely:
Yes. But what I was wondering is, should this be the default behavior in forem?

parndt:
What alternative are you proposing?

daniely:
https://github.com/radar/forem/blob/master/app/models/forem/ability.rb#L44

change from

``` ruby
user.can_reply_to_forem_topic?(topic)
```

to this

``` ruby
can?(:read, topic.forum) && user.can_reply_to_forem_topic?(topic)
```

then it would be similar to the permissions for creating topics

parndt:
Can you send this as a pull request please?

daniely:
yes of course, just wanted to first make sure it was something you guys wanted :)

i'll write and test and submit after work today.

parndt:
@radar confirm?

radar:
Yes. A user should only be able to reply to a topic if they have read permission on the forum. Please do submit a PR for that today.

PR #446, opened 2013-07-17
check forum read permissions when user replies to topic/post

daniely:
fix for #445

i usually try to stay away from `any_instance` but here it was the most succinct way i could do it. if there's a better way to do this i'm all ears.

test was written at the controller level since it's about trying to bypass permissions by making a clever http POST. a feature test didn't make sense since presumably the user would not have access to links for replying to a post.

### DIFF
--- a/app/models/forem/ability.rb
+++ b/app/models/forem/ability.rb
@@ -41,7 +41,7 @@ module Forem
       end
 
       can :reply, Forem::Topic do |topic|
-        user.can_reply_to_forem_topic?(topic)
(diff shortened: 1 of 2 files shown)


radar:
Thank you @daniely. I really appreciate you taking the time to make this contribution.

I've made a small modification to your patch to use `controller.current_user.stub` rather than `any_instance` and that seems to work.

daniely:
@radar welcome! and thanks for showing me a better way to do that stub.

PR #450, opened 2013-07-26
bootstrap-sass isn't needed by default

jugyo:
bootstrap-sass isn't needed by default. It should be took care by theme such as forem-bootstrap.
So this change will remove dependency to bootstrap-sass.


AndorChen:
:+1:

radar:
Thank you.

Issue #448, opened 2013-07-26
There is no locale data for 'forem.admin.forums.index.no_category'


jugyo:
It's needed at here: https://github.com/radar/forem/blob/master/app/views/forem/admin/forums/index.html.erb#L7

To reproduce:
- create a category
- create topic in the category
- delete the category
- go to /admin/forums
- "en forem.admin.forums.index.no_category" is shown

radar:
@jugyo Thank you for pointing this out. Could you please submit a patch to fix this problem? That would be awesome.

jugyo:
BTW, Is it OK that we can delete category which still have topic?

radar:
Forums are the container that contain topics, not categories. Deleting categories should not delete topics.

jugyo:
Sorry, I meant forum, not topic.

radar:
Forums can be reassigned to categories, so no they won't be deleted when a category is deleted.

jugyo:
That makes sense. Thanks.

daniely:
close?

jugyo:
LGTM, thanks!

PR #455, opened 2013-08-13
add locale data for forem.admin.forums.index.no_category

daniely:
Fix for #448. Can anyone recommend a good test to write for this? A view test?


daniely:
FYI, I tested this locally to make sure it works.

parndt:
I think this is fine without a test; thanks!

PR #386, opened 2013-02-24
DB Optimizations

phlipper:
This branch contains some database improvements including eliminating n+1 queries and adding some cache_counter columns.

These changes eliminate 100's of queries per request for large installations. 

More work remains for future pull requests.

PR #462, opened 2013-08-17
Remove n+1

daniely:
some db optimizations mentioned in #386

most of the legwork was done by @phlipper. i simply did some cleanup and applied it to the newest master branch.

i'm splitting up the n+1 and counter cache into separate commits

### DIFF
--- a/app/controllers/forem/categories_controller.rb
+++ b/app/controllers/forem/categories_controller.rb
@@ -1,6 +1,14 @@
 module Forem
   class CategoriesController < Forem::ApplicationController
+    before_filter :find_category, :only => :show
+
     helper 'forem/forums'
     load_and_authorize_resource :class => 'Forem::Category'
+
+    private
+
+    def find_category
+      @category = Forem::Category.includes(:forums => :category).find(params[:id])
+    end
   end
 end
--- a/app/controllers/forem/forums_controller.rb
+++ b/app/controllers/forem/forums_controller.rb
@@ -4,7 +4,7 @@ module Forem
     helper 'forem/topics'
 
     def index
-      @categories = Forem::Category.all
+      @categories = Forem::Category.includes(:forums => [:category, :moderators]).all
     end
 
     def show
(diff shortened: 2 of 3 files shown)


app/controllers/forem/forums_controller.rb
```diff
@@ -4,7 +4,7 @@ class ForumsController < Forem::ApplicationController
     helper 'forem/topics'
 
     def index
-      @categories = Forem::Category.all
+      @categories = Forem::Category.includes(:forums => [:category, :moderators]).all
```
parndt: This means we don't need the `.all` anymore

app/controllers/forem/categories_controller.rb
```diff
@@ -1,6 +1,14 @@
...
   class CategoriesController < Forem::ApplicationController
+    before_filter :find_category, :only => :show
+
     helper 'forem/forums'
     load_and_authorize_resource :class => 'Forem::Category'
+
+    private
+
+    def find_category
+      @category = Forem::Category.includes(:forums => :category).find(params[:id])
```
radar, line 11: I don't understand the `includes` here. Why is that necessary if all the forums have the same category? Shouldn't it just be `includes(:forums)`?

app/controllers/forem/forums_controller.rb
```diff
@@ -4,7 +4,7 @@ class ForumsController < Forem::ApplicationController
     helper 'forem/topics'
 
     def index
-      @categories = Forem::Category.all
+      @categories = Forem::Category.includes(:forums => [:category, :moderators])
```
radar, line 7: I don't think the `category` should be here in the includes, either.

app/controllers/forem/forums_controller.rb
```diff
@@ -16,7 +16,8 @@ def show
         @forum.topics.visible.approved_or_pending_review_for(forem_user)
       end
 
-      @topics = @topics.by_pinned_or_most_recent_post.page(params[:page]).per(Forem.per_page)
+      @topics = @topics.by_pinned_or_most_recent_post.includes(:forum, :user)
```
radar, line 19: Why does `forum` need to be included here?

app/controllers/forem/topics_controller.rb
```diff
@@ -92,7 +92,7 @@ def find_forum
     end
 
     def find_posts(topic)
-      posts = topic.posts
+      posts = topic.posts.includes(:user, :reply_to, :topic)
```
radar, line 95: Why does `topic` need to be included here?

PR #463, opened 2013-08-17
see if updating rspec fixes travis builds

daniely:
just a test. not sure how else to fire off travis builds


daniely:
upgrading rspec didn't work

PR #466, opened 2013-08-20
test again to see if this fixes jruby builds

daniely:


daniely:
works. closing this and cherry-picking into an existing PR

PR #468, opened 2013-08-26
fix some links to forums page for consistency

jugyo:
For consistency, It would be better that you use forem.root_path (/) instead of forem.forums_path (/forums/).


radar:
I agree!

jugyo:
thanks :smile:

Issue #470, opened 2013-09-03
Error when installing and further finders error


link-er:
Hello!
I'm trying to use rails4 branch. I installed gem and run install generator. Here what I got:

``` bash
rails g forem:install
I, [2013-09-03T15:29:38.358409 #1802]  INFO -- : ** [Raven] Raven 0.6.0 ready to catch errors
Copying over Forem migrations...
What is your user class called? [User] User
What is the current_user helper called in your app? [current_user] current_user
Defining forem_user method inside ApplicationController...
      insert  app/controllers/application_controller.rb
Adding forem initializer (config/initializers/forem.rb)...
      create  config/initializers/forem.rb
Running rake db:migrate
Creating default forum and topic
/home/lina/.rvm/gems/ruby-2.0.0-p195/gems/activemodel-4.0.0/lib/active_model/attribute_methods.rb:436:in `method_missing': undefined method `forem_moderate_posts?' for #<User:0x00000006dd6b60> (NoMethodError)
    from /home/lina/.rvm/gems/ruby-2.0.0-p195/gems/activerecord-4.0.0/lib/active_record/attribute_methods.rb:131:in `method_missing'
```

I thought that it's not very critical for engine working, so I continued, added route and started application. But then, when I tried to get to default forum's page, I got next error:

```
ActiveRecord::RecordNotFound in Forem::ForumsController#show
Couldn't find Forem::Forum with id=default
```

The same with categories.
Is these problems are bugs or am I doing something wrong?

parndt:
Likely bugs.. the error about not being able to find with id=default is likely FriendlyId's API change.

link-er:
adding  `:find_by => :slug` in `load_and_authorize_resource :class => 'Forem::Forum', :only => :show`
 in controller cures the problem (the same in categories controller)
shall I make pull request or you can help?

link-er:
Error on install seems to be happened because nothing was wrote to my User model
As I can see in lib/forem.rb file, there is a method for decorating model, but... smth went wrong

P.S. can you please tell me what else has to be done on generator install, maybe there is anything else missing

kekiefer:
Regarding the id=default bug here, you should use nukturnal/cancan, which fixes the friendly_id 5 issue as @parndt mentioned. You could also fix the issue by modifying the forem model finders (https://github.com/kekiefer/forem/commit/c7717ffab3bb6a8a654c27adfc4f8bb22bea24d7).

Senjai:
[...]

PR #471, opened 2013-09-10
FriendlyId 5 compatible finders -- restore FriendlyId 4-style finders by...

kekiefer:
... using the :finders addon

From the FiendlyId 5 docs (rails 4) --

> Finders are no longer overridden by default. If you want to do friendly finds, you must do Model.friendly.find rather than Model.find. You can however restore FriendlyId 4-style finders by using the :finders addon:
> 
>   friendly_id :foo, use: :slugged # you must do MyClass.friendly.find('bar')
>   # or...
>   friendly_id :foo, use: [:slugged, :finders] # you can now do MyClass.find('bar')

This commit enables the :finders addon, fixing problems with finder errors (https://github.com/radar/forem/issues/470), e.g.:

```
ActiveRecord::RecordNotFound in Forem::ForumsController#show
Couldn't find Forem::Forum with id=default
```

### DIFF
--- a/app/models/forem/category.rb
+++ b/app/models/forem/category.rb
@@ -3,7 +3,7 @@ require 'friendly_id'
 module Forem
   class Category < ActiveRecord::Base
     extend FriendlyId
-    friendly_id :name, :use => :slugged
+    friendly_id :name, :use => [:slugged, :finders]
 
     has_many :forums
     validates :name, :presence => true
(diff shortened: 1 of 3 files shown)


kekiefer:
Sorry, all, I missed the root cause here fixed through nukturnal/cancan, specifically https://github.com/nukturnal/cancan/commit/a606378912e75d6b62f8257ffa5b4c5f2023760b

If for some reason you don't want to use nukturnal/cancan (and you don't use friendly_id in your project for other stuff), this commit will work around the issue.

PR #473, opened 2013-09-12
Identify hidden topics in administrator's view

tigercat:
It's helpful for administrators to see which topics are hidden and which are not when looking at a forum.  Otherwise,
one has to click into each topic to see whether it is hidden.

Added a span with classes "hidden" and "icon" to the topic partial, which is added to the topic view when the topic is hidden. Using this span, one can create an icon in a topics view to identify hidden topics.

### DIFF
--- a/app/views/forem/topics/_topic.html.erb
+++ b/app/views/forem/topics/_topic.html.erb
@@ -6,6 +6,9 @@
     <% if topic.pinned? %>
       <span class='pin icon'></span>
     <% end %>
+    <% if topic.hidden? %>
+      <span class='hidden icon'></span>
+    <% end %>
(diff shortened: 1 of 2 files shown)


parndt:
So to confirm we're not supplying an icon?

tigercat:
As far as I know, there are no icons of this type included with the 
forem gem (and the current version provides classes for icons for locked 
and pinned topics - see _topic.html.erb).  For this reason, I didn't try 
to supply a "hidden" icon.  These sort of icons currently come with 
theme gems (see forem-theme-base), so that would be the the place to add 
a "hidden" icon.

radar:
Looks good to me. Thank you :)

PR #475, opened 2013-09-17
CHANGED: corrected japanese localization for forum view counts

TDaulat:


radar:
Added to master and rails4. Thank you.

PR #480, opened 2013-09-30
Add moderation topics

tigercat:
Added a view of topics that need moderating to the moderation index page.

Thought about different ways to do this and decided to use a list of links to the topic moderation pages.  Seemed cleaner to read than listing all the topics moderation views explicitly (and easier to do).

Added tests.


tigercat:
Changes are what I want, but want to clean up extraneous files in commit list (should only be 2, not 7).  Will resubmit when I straighten list out.

PR #484, opened 2013-09-30
Added a list of topics to the moderation view

tigercat:
The moderation index only showed posts that needed moderation, but not topics.  Added a list of topics that need moderation.

I went with a list of links rather than complete topics, as complete topics are visually pretty large and would make the view less clear.  A list of complete topics is also more complex to implement.

There is another earlier commit (included here) that used forms in the view, but I overwrote it, because I decided links are clearer.

### DIFF
--- a/app/views/forem/moderation/index.html.erb
+++ b/app/views/forem/moderation/index.html.erb
@@ -11,3 +11,17 @@
   </div>
   <%= submit_tag "Moderate", :class => "btn btn-primary" %>
 <% end %>
+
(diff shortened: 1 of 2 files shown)


parndt:
Thanks for this, could you please squash your commits into a single commit so that the git log can be a little clearer regarding this feature? :heart:

tigercat:
Squashing is done.

parndt:
Dankeschön..  @radar @knewter wdyt?

radar:
Looks fine to me!

PR #485, opened 2013-09-30
Added ability to control whether users can delete posts.

tigercat:
Similar to the currently existing ability to control whether users
can edit posts.

Fixed post view, so buttons reflect whether ability exists or not.

Added tests.


radar:
Thanks! Added to rails4 and master.

PR #487, opened 2013-10-06
Fixed russian pluarization

NOX73:
Moderation page raised I18n::InvalidPluralizationData  exception for russian language, if count more then 2.

```
I18n::InvalidPluralizationData can not be used with :count => 3
```


radar:
Thanks!

PR #488, opened 2013-10-09
fix friendly_id

karquelf:
Hello,

I am using your gem with rails4.

Here is a small fix for the branch rails4 in order to work with friendly_id 5.0 https://github.com/norman/friendly_id

It avoids errors like "Couldn't find Forem::Category with id=general" for the url : http://localhost:3000/forums/categories/general

Best regards


radar:
This patch has already been made to the rails4 branch.

Issue #325, opened 2012-10-27
forem:install generator fails on refinerycms


eostrom:
I installed Refinery CMS:

``` sh
gem install refinerycms
refinerycms reforemery
cd reforemery
rails server
```

Fired it up and created the initial user.

Then I installed forem, as per the README, by adding to Gemfile:

``` ruby
gem 'forem', :git => "git://github.com/radar/forem.git"
```

And at the command line:

``` sh
$ rails g forem:install
Copying over Forem migrations...
What is your user class called? [User] Refinery::User
What is the current_user helper called in your app? [current_user] current_refinery_user
Defining forem_user method inside ApplicationController...
      insert  app/controllers/application_controller.rb
Adding forem initializer (config/initializers/forem.rb)...
      create  config/initializers/forem.rb
Running rake db:migrate
Creating default forum and topic
/Users/eostrom/.rvm/gems/ruby-1.9.2-p290@iowl/gems/activemodel-3.2.6/lib/active_model/attribute_methods.rb:407:in `method_missing': undefined method `forem_needs_moderation?' for #<Refinery::User:0x007ff4ea951690> (NoMethodError)
    from /Users/eostrom/.rvm/gems/ruby-1.9.2-p290@iowl/gems/activerecord-3.2.6/lib/active_record/attribute_methods.rb:149:in `method_missing'
    from /Users/eostrom/.rvm/gems/ruby-1.9.2-p290@iowl/gems/activesupport-3.2.6/lib/active_support/core_ext/object/try.rb:36:in `try'
    from /Users/eostrom/.rvm/gems/ruby-1.9.2-p290@iowl/bundler/gems/forem-66e2f512d0f0/app/models/forem/post.rb:116:in `skip_pending_review'
    from /Users/eostrom/.rvm/gems/ruby-1.9.2-p290@iowl/gems/activesupport-3.2.6/lib/active_support/callbacks.rb:417:in `_run__280411519566598945__create__954061338601118689__callbacks'
    from /Users/eostrom/.rvm/gems/ruby-1.9.2-p290@iowl/gems/activesupport-3.2.6/lib/active_support/callbacks.rb:405:in `__run_callback'
    from /Users/eostrom/.rvm/gems/ruby-1.9.2-p290@iowl/gems/activesupport-3.2.6/lib/active_support/callbacks.rb:385:in `_run_create_callbacks'
[...]

PR #496, opened 2013-10-30
Added attr_accessible to Forem Category on rails4 branch

mijoharas:
This broke the installer for me [#325](https://github.com/radar/forem/issues/325) as `Forem::Category.create(:name => 'General')` in `db/seeds.rb` would not run due to rails 4 strong parameters. Fix already exists on master.

### DIFF
--- a/app/models/forem/category.rb
+++ b/app/models/forem/category.rb
@@ -7,6 +7,7 @@ module Forem
 
     has_many :forums
     validates :name, :presence => true
+    attr_accessible :name
 
     def to_s
       name
--- a/app/models/forem/forum.rb
+++ b/app/models/forem/forum.rb
@@ -16,6 +16,8 @@ module Forem
 
     validates :category, :name, :description, :presence => true
 
+    attr_accessible :category_id, :title, :name, :description, :moderator_ids
+
     alias_attribute :title, :name
 
     # Fix for #339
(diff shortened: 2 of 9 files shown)


mijoharas:
nevermind, just realised that attr_accessible is missing on most of the models on rails4 branch, they appear to be there on the master branch, I if we pull those commits in then this should allow the installation via `rails g forem:install` on this branch.

mijoharas:
added the rest of them.

radar:
This is not necessary because in Rails 4 land we are using strong_parameters instead. Please remove the protected-attributes gem from your Gemfile, which is causing this problem.

mijoharas:
Ahhh, thanks, I was wondering about the atypical behaviour, hadn't noticed that I'd included the protected-attributes when I'd installed refinery. Thanks.

parndt:
Ah, you may have issues with Refinery. Nobody's taken the time to upgrade that yet to not use attr_accessible. cc @ugisozols

mijoharas:
ok, I may dive into that and clean up what needs cleaning up in refinery soon if there are many problems, (I'll submit a pull request with the fix if I make it). In the meantime I've got a branch with the changes in forem that allow me to use the protected-attributes gem if needed. Thanks.

PR #497, opened 2013-11-06
Bump select2-rails dependency

raulpopadineti:
I wasn't able to bundle the 'forem' gem on 'spree' 2-1-stable because of the fixed '3.4.3' version. After the bump, everything worked fine.


radar:
Thanks!

Issue #495, opened 2013-10-21
Two related (?) issues - undefined method `forem_moderate_posts?; undefined local variable or method `name'


philiprhoades:
People,

Firstly, thanks for this software!

I have looked through old issues but can't see this covered:

This: https://github.com/jayshepherd/devise_example

works fine. I then install this: https://github.com/radar/forem

but the first time I run: rails g forem:install

I get:

Running rake db:migrate
/home/phil/.gem/ruby/gems/activemodel-4.0.0/lib/active_model/attribute_methods.rb:436:in `method_missing': undefined method`forem_moderate_posts?' for #<User
:0x00000004e6d698> (NoMethodError)

but when I run it again it appears to finish OK. Logging in as the test user and looking at: /forums

works OK but when I add:

def to_s
  name
end

to the user model at /forums I get:

Showing /home/phil/.gem/ruby/bundler/gems/forem-f34fb3d65583/app/views/forem/forums/_forum.html.erb where line #8 raised:

undefined local variable or method `name' for #User:0x00000003c72f60

Fixes appreciated!

Regards,
Phil.

radar:
Hi @philiprhoades,

Could you please show me the complete output from `rails g forem:install`? I don't see the prompts in the message you posted and I'm interested to see what values you put in for that.

Thanks!

philiprhoades:
radar,

I just used the defaults:

```
[DEVISE] Devise.use_salt_as_remember_token is deprecated and has no effect. Please remove it.
Copying over Forem migrations...
What is your user class called? [User] What is the current_user helper called in your app? [current_user] Defining forem_user method inside ApplicationController...
[1m[32m      insert[0m  app/controllers/application_controller.rb
Adding forem initializer (config/initializers/forem.rb)...
[1m[32m      create[0m  config/initializers/forem.rb
Running rake db:migrate
/home/phil/.gem/ruby/gems/activemodel-4.0.0/lib/active_model/attribute_methods.rb:436:in `method_missing': undefined method `forem_moderate_posts?' for #<User:0x00000004e6d698> (NoMethodError)
    from /home/phil/.gem/ruby/gems/activerecord-4.0.0/lib/active_record/attribute_methods.rb:131:in `method_missing'
    from /home/phil/.gem/ruby/bundler/gems/forem-f34fb3d65583/app/models/forem/post.rb:112:in `skip_pending_review'
[...]

PR #498, opened 2013-11-18
Mention the fact that Devise does not provide 'name' column by default

bodrovis:
Partially fixes https://github.com/radar/forem/issues/495.

Some users (like me :)) may forget about the fact that Devise does not provide `name` column by default and such error raises: `undefined local variable or method`name' for #User:0x00000003c72f60`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,9 @@ def to_s
 end
 ```
 
+Please note that if you are using Devise, User model does not have `name` column by default,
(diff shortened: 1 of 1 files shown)


radar:
Hi @bodrovis, can you please remove the russian translation changes from this PR?

bodrovis:
@radar Done!

PR #499, opened 2013-11-20
Ability to remove users from groups

bodrovis:
As far as I understand there is no way to remove user from the group after you added them (of course it can be done via console). I have started to implement this functionality and it does work. I had to specify in routes:

``` ruby
resources :members, only: [:destroy] do
```

because otherwise the following error appeared in my console: 

```
AbstractController::ActionNotFound - The action 'show' could not be found for Forem::Admin::MembersController:
```

Am I missing some concept here?

I cannot use tests, because running `bundle exec rake forem:dummy_app` gives me an error (will open a separate issue). It would be nice if someone can help me with tests.


.gitignore
```diff
@@ -12,3 +12,4 @@ spec/dummy
 *.rbc
 *.tmproj
 Gemfile.lock
+.idea/
```
parndt: This belongs in your `~/.gitignore`

bodrovis: Sorry, it's just RubyMine config files. Makes me sick removing it every time :). Still removed .idea from gitignore.

bodrovis:
I am sorry, will replace this pull request with anoter one https://github.com/radar/forem/pull/503 because I really messed with Git here and it became cluttered.

Issue #500, opened 2013-11-20
Not being able to run tests


bodrovis:
I tried to run tests when adding some functionality, but no luck. Running

``` ruby
bundle exec rake forem:dummy_app
bundle exec rspec spec
```

gives me these errors: http://pastebin.com/CCLpY993

I have ruby 1.9.3 on my machine.

radar:
Can you please show me the `Gemfile` that you're using? According to the error message, you don't have the sqlite3 adapter installed. This is peculiar.

bodrovis:
``` ruby
source "https://rubygems.org"

gem 'forem-theme-base', :git => "git://github.com/radar/forem-theme-base", :branch => "master"

gemspec

gem 'pry-rails'
gem 'pry-nav'

platforms :jruby do
  gem "activerecord-jdbc-adapter", :require => false
end

group :test do
  platforms :ruby do
    gem "forem-redcarpet", :github => "radar/forem-redcarpet", :branch => "master"
    gem "mysql2"
    gem "pg"
    gem "sqlite3"
  end

  platforms :jruby do
    gem "activerecord-jdbcmysql-adapter", :require => false
    gem "activerecord-jdbcpostgresql-adapter", :require => false
    gem "activerecord-jdbcsqlite3-adapter", :require => false
    gem "forem-kramdown", :github => "phlipper/forem-kramdown", :branch => "master"
  end
end

if RUBY_VERSION < '1.9.2'
  gem 'nokogiri', '~> 1.5.9'
end
```

Maybe I know what is causing this problem. It is because we have `platforms :ruby do` where `sqlite3` is present, but I am currently on Windows and so gems from this group are not being loaded.

bodrovis:
I changed that line to `platforms :ruby, :mswin do`, but now when trying to run tests I keep getting 

```
git://github.com/radar/forem-redcarpet.git (at master) is not checked out. Please run bundle install
```

Running `bundle install` or `bundle install --no-deployment` does not help :(

Updated:

```
bundle pack
bundle install --path vendor/cache
```

didn't help either.

radar:
I don't know how this could be happening :( I would need a way to reproduce this exactly on my computer if I was going to debug it.

bodrovis:
@radar Wouldn't cloning the repo to your PC and modifying the line to `platforms :ruby, :mswin do` do the trick? I will try to research more today on this issue.

bodrovis:
[...]

PR #503, opened 2013-11-21
Ability to remove users from groups

bodrovis:
As far as I understand there is no way to remove user from the group after you added them (of course it can be done via console). I have started to implement this functionality and it does work. I had to specify in routes:

``` ruby
resources :members, only: [:destroy] do
```

because otherwise the following error appeared in my console: 

```
AbstractController::ActionNotFound - The action 'show' could not be found for Forem::Admin::MembersController:
```

Am I missing some concept here?

I cannot use tests, because running `bundle exec rake forem:dummy_app` gives me an error (will open a separate issue). It would be nice if someone can help me with tests.


radar:
Patch looks good to me. Thanks @bodrovis.

radar:
I would really like to see some tests for this too.

bodrovis:
@radar Yes, I will try to add tests after we solve this one https://github.com/radar/forem/issues/500 :)

PR #504, opened 2013-11-21
Reversible migrations

mauriciopasquier:
I had to rollback forem migrations for some reason and found these bugs.. Not really critical, but a bug is a bug :)


radar:
Thanks @mauriciopasquier. I've added this to master + rails4.

mauriciopasquier:
thanks :)

PR #505, opened 2013-11-25
Use Kaminari configuration if present

mauriciopasquier:
First of all, this is not a kaminari vs will_paginate patch but it's [related to](https://github.com/amatsuda/kaminari/issues/162#ref-commit-3cf929f). In that commit amatsuda introduces some configuration variables for kaminari. I'm using them like

```
Kaminari.configure do |config|
  config.page_method_name = :pagina
  config.param_name = :pagina
end
```

not because of conflict, but for i18n issues. And forem is still calling `page` and using `params[:page]`.
This patch uses that configuration if present, and defaults to `page` if not.

I can add some tests if this approach is welcome.

### DIFF
--- a/app/controllers/forem/application_controller.rb
+++ b/app/controllers/forem/application_controller.rb
@@ -9,6 +9,18 @@ class Forem::ApplicationController < ApplicationController
     Forem::Ability.new(forem_user)
   end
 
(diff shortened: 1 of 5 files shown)


radar:
Patch looks good to me.

PR #507, opened 2013-11-28
Windows platform should be specified in Gemfile to install required gems...

bodrovis:
... in testing env (fixes this issue https://github.com/radar/forem/issues/500). We also shouldn't use `"forem-redcarpet"` directly from master branch as new version was released two days ago.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,8 +12,8 @@ platforms :jruby do
 end
 
 group :test do
-  platforms :ruby do
-    gem "forem-redcarpet", :github => "radar/forem-redcarpet", :branch => "master"
+  platforms :ruby, :mingw do
+    gem "forem-redcarpet"
     gem "mysql2"
     gem "pg"
     gem "sqlite3"


Gemfile
```diff
@@ -12,8 +12,8 @@ platforms :jruby do
 end
 
 group :test do
-  platforms :ruby do
-    gem "forem-redcarpet", :github => "radar/forem-redcarpet", :branch => "master"
+  platforms :ruby, :mingw do
```
radar, line 15: This means that these gems will only be installed on Windows. This is bad.

bodrovis, line 15: Why so? It means that these will be installed on Windows AND on other platforms as per this document: http://bundler.io/v1.2/man/gemfile.5.html. I've put more info on this comment: https://github.com/radar/forem/issues/500#issuecomment-29050211

bodrovis, line 15: Could you please check it on your PC (i presume, you are not working on Windows)?

radar, line 15: Oh I see, you're combining the platforms so that allows for all of them. I use Macs exclusively for development, but I do have a Windows PC that I occasionally use for gaming that also has Ruby installed, so I could check there.

PR #509, opened 2013-12-05
Arabic Translation

moh-alsheikh:
not to much just added arabic translation.


radar:
Thank you @moh-alsheikh. I've added this to master + rails4 now.

Issue #513, opened 2013-12-11
Post.text is being rendered without paragraphing


avitus:
We're using Rails 4.0.2 and

```
gem 'forem',       :github => "radar/forem", :branch => "rails4"
```

The text in the database is as follows:

```
"This is the first paragraph.\r\n\r\nThis is the second paragraph."
```

But it is rendering without any paragraphing at all. We recently updated the Forem gem and prior to that everything was ok. I suspect it is related to this commit:

Remove simple_format from formatting_helper: https://github.com/radar/forem/commit/2bc42231ea0a8d61968e5105a56b5a72c449c68a

bodrovis:
Seems so. You could also take a look here https://github.com/radar/forem/issues/512 and here https://github.com/radar/forem/issues/359

bodrovis:
You could try to use some formatter for the text (https://github.com/radar/forem/wiki/Formatters), for example https://github.com/phlipper/forem-kramdown/. At least I launched console in my test Forem app and tried to format your text with kramdown and paragraphs appeared. However an issue with sanitization could rise which is described here https://github.com/radar/forem/issues/512. I've bumped into it a couple of weeks ago and had to make a temporal fix: https://github.com/bodrovis/forem/commit/72d592f3b420958163da0b3a5958935d97a30a99

avitus:
@bodrovis Thanks for the tip. I used the Textile one and that seems to have fixed the problem. I had assumed that using a Formatter was optional but extracting the functionality makes sense.

Maybe the installation part of the Readme should call that out if that's the way forward.

bodrovis:
@avitus It was optional until `simple_format` was removed :) I will make a pull request to clarify this issue in readme

PR #514, opened 2013-12-12
Notice about formatters and bump will_paginate to 3.0.5 in Readme

bodrovis:
Per this issue: https://github.com/radar/forem/issues/513

Probably we should remind users to specify formatter however I am not sure wheter this should be done in Readme or in Wiki.


radar:
This has been added to rails4.

PR #518, opened 2013-12-22
Cannot quote deleted post

alexcwatt:
Clicking "Quote" on a post that has since been deleted throws the following error in an application using forem:

```
Error Message:
NoMethodError: undefined method `text' for nil:NilClass

Where:
forem/posts#new
[GEM_ROOT]/bundler/gems/forem-ef6a4679928c/app/controllers/forem/posts_controller.rb, line 17
```

I wasn't sure how to add a new flash notice... it's just English. (I'm not sure if adding a translation for it to the `en.yml` file but not the others would cause problems.)

This pull request is intended for the master branch as well.

### DIFF
--- a/app/controllers/forem/posts_controller.rb
+++ b/app/controllers/forem/posts_controller.rb
@@ -17,7 +17,7 @@ module Forem
       if params[:quote] && @reply_to_post
         @post.text = view_context.forem_quote(@reply_to_post.text)
       elsif params[:quote] && !@reply_to_post
-        flash[:notice] = t("forem.post.cannot_quote")
+        flash[:notice] = t("forem.post.cannot_quote_deleted_post")
         redirect_to [@topic.forum, @topic]
       end
     end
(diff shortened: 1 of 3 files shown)


alexcwatt:
Thanks @bodrovis -- just updated it.

alexcwatt:
@radar Is there anything else I should do with this? Thanks for all your work on forem.

app/controllers/forem/posts_controller.rb
```diff
@@ -14,7 +14,12 @@ def new
       @post = @topic.posts.build
       find_reply_to_post
 
-      @post.text = view_context.forem_quote(@reply_to_post.text) if params[:quote]
+      if params[:quote] && @reply_to_post
+        @post.text = view_context.forem_quote(@reply_to_post.text)
+      elsif params[:quote] && !@reply_to_post
+        flash[:notice] = t("forem.post.cannot_quote")
```
radar: I would probably call this `cannot_quote_deleted_post`.

radar:
@alexcwatt Thanks! The patch looks good to me. 

I've been on holidays for a couple of weeks and have been playing email catch up end of last and this week, which is why I hadn't replied to it yet. Back now though!

app/controllers/forem/posts_controller.rb
alexcwatt: Good idea. Just updated with c7c5a501f82ac86bb13b2d2ca02ff57624b1289c.

radar:
Added to rails4 and master. Thanks @alexcwatt.

PR #521, opened 2014-01-09
Fix formatting blockquote test

alexcwatt:
The second test ("renders blockquote without p tag wrapping") was failing when another post existed before it, so I updated the test to look only at the last post.

You can check that the test was failing with `bundle exec rspec --seed 3056` before this and that it now passes.


radar:
Thanks @alexcwatt! I can confirm this breakage and that the PR fixes it. Excellent work.

alexcwatt:
You're welcome! Hoping that the Travis builds can be all green again soon...

PR #522, opened 2014-01-13
Use POST for subscribing/unsubscribing from forum topics.

KL-7:
Use POST request with `authenticity_token` instead of GET to protect topics (un)subscribing process from CSRF attack.


radar:
Thank you very much @KL-7! I've added this to master and rails4 now.

torasakura:
that's nice, but the unsubscribe link in topic reply email is not working. i get a No route matches [GET] error for unsubscribe url.

radar:
@torasakura Please submit a patch to fix this. `unsubscribe` should work with either a `POST` or a `GET`.

PR #523, opened 2014-01-19
Add missing translation keys and lookups

ascendbruce:
This is a simplify version of #406, which contains only "fix missing key and missing lookups"


radar:
Thanks @ascendbruce. I've added this to master and rails4 now.

PR #536, opened 2014-02-20
Fixed a regression introduced about by_pinned_or_most_recent_post

HollowerMan:
I noticed a change on by_pinned_or_most_recent_post I could not explain and that caused a strange behavior on my pinned topics.
I just reverted to what the code was on master, and everything works again.

### DIFF
--- a/app/models/forem/topic.rb
+++ b/app/models/forem/topic.rb
@@ -52,8 +52,9 @@ module Forem
       end
 
       def by_pinned_or_most_recent_post
-        order('forem_topics.last_post_at DESC').order('forem_topics.pinned DESC')
-        #order('forem_topics.id')
+	order('forem_topics.pinned DESC').
+        order('forem_topics.last_post_at DESC').
+        order('forem_topics.id')
       end
 
       def pending_review
--- a/spec/models/topic_spec.rb
+++ b/spec/models/topic_spec.rb
@@ -47,6 +47,13 @@ describe Forem::Topic do
     end
   end
 
+  describe ".by_pinned_or_most_recent_post" do
+    it "should show topics by pinned then by most recent post" do
+      ordering = Forem::Topic.by_pinned_or_most_recent_post.order_values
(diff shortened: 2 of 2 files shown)


radar:
@HollowerMan Do you have a description of the bug that you were seeing? Having a regression test that would demonstrate this bug would be really great.

HollowerMan:
@radar First, the bug.
My pinned topics are simply not displayed on top of the other topics in my app forum. They are just ordered by date like the others.

Then, I tried to add a regression test, but there is a behavior I cannot explain.
My patched `Forem::Topic.by_pinned_or_most_recent_post` is defined to:

```
Forem::Topic.order('forem_topics.pinned DESC').order('forem_topics.last_post_at DESC').order('forem_topics.id')
```

So when I do:

```
Forem::Topic.by_pinned_or_most_recent_post.order_values
```

I expect the following output:

```
["forem_topics.pinned DESC", "forem_topics.last_post_at DESC", "forem_topics.id"]
```

This is working in my application console using the patch I provided here.

But this very test fails with `bundle exec rspec spec`:

```
    Forem::Topic.by_pinned_or_most_recent_post should show topics by pinned then by most recent post
    Failure/Error: ordering.should == ["forem_topics.pinned DESC", "forem_topics.last_post_at DESC", "forem_topics.id"]
      expected: ["forem_topics.pinned DESC", "forem_topics.last_post_at DESC", "forem_topics.id"]
        got: ["forem_topics.id", "forem_topics.last_post_at DESC", "forem_topics.pinned DESC"] (using ==)
    # ./spec/models/topic_spec.rb:53:in `block (3 levels) in <top (required)>'
```

Please let me know if I can provide more details about those issues.

radar:
@HollowerMan I think this was already tested within `spec/features/forums_spec.rb`, and I've seen that test failing. I pulled in your PR and it has seemingly fixed that test, so thanks. This patch looks great :)

Issue #474, opened 2013-09-16
Moderating Topics Created by Unapproved Users


tigercat:
If an unapproved user creates a topic, they get a warning from the system that the topic is pending review.  However, when an administrator clicks the "Moderation Tools" button, forem says "No posts".  But there is a post for a new topic created by the unapproved user. 

Looking at the code for the ModerationController:

``` ruby
def index
  @posts = forum.posts.pending_review.topic_not_pending_review
  @topics = forum.topics.pending_review
end
```

One can see `@posts` doesn't include pending topics.  These are stored in `@topics`.

However, the moderation/index view (simplified) is:

``` erb
<%= t('.title', :forum => forum) %>

<%= t('posts_count', :count => @posts.count, :scope => 'forem.general') %>

<%= form_tag forem.forum_moderate_posts_url(forum), :method => :put do %>
  <div id='posts'>
    <% @posts.limit(25).group_by(&:topic).each do |topic, posts| %>
      <%= forem_emojify(topic.forum.title) %> &raquo; <%= forem_emojify(topic.subject) %>
      <%= render posts, :mass_moderation => true %>
    <% end %>
  </div>
  <%= submit_tag "Moderate", :class => "btn btn-primary" %>
<% end %>
```

There is no mention of `@topics`. 

What is the plan here?  Was the moderation/index view meant to include `@topics`?  If we don't display `@topics`, how can administrators easily check for new topics that are pending review?

radar:
Hi @tigercat.

Please remember to format code with triple backticks. This makes it easier for me to see what's code and what's text. I've edited your original post here and you can see what I mean by clicking "Edit" again.

radar:
Yes, the moderation view was supposed to contain topics. This has obviously been overlooked. Would you please be able to submit a patch that adds a list of topics to this view please?

tigercat:
Sorry about the formatting.  Will do it in the future.

I'll look into enhancing the view for topics.

radar:
@tigercat any updates on this?

bodrovis:
<s>I was not able to reproduce this using `rails3` branch. I have two users: an administrator and a ordinary user with `forem_state` set to `pending_review`. I add a post as an ordinary user, then login as an admin, open moderation tools and do see this topic and can approve it. Am I missing something or was it fixed somewhere else?</s>

bodrovis:
Oops sorry I still see "no posts" message.

radar:
Fixed by #545.

PR #545, opened 2014-03-19
Showing unapproved posts within unapproved topics

bodrovis:
Should fix https://github.com/radar/forem/issues/474 Tests do pass.

### DIFF
--- a/app/views/forem/moderation/index.html.erb
+++ b/app/views/forem/moderation/index.html.erb
@@ -9,14 +9,13 @@
       <%= render posts, :mass_moderation => true %>
     <% end %>
   </div>
-  <%= submit_tag t('forem.posts.moderation.moderate'), :class => "btn btn-primary" %>
 <% end %>
 
 <h3><%= t('topics_count', :count => @topics.count, :scope => 'forem.forum') %></h3>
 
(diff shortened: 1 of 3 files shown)


app/views/forem/moderation/index.html.erb
```diff
@@ -9,14 +9,13 @@
       <%= render posts, :mass_moderation => true %>
     <% end %>
   </div>
-  <%= submit_tag t('forem.posts.moderation.moderate'), :class => "btn btn-primary" %>
```
radar, line 12: I don't understand why this button has been removed.

bodrovis, line 12: Post is being rendered with its own submit button, so for 2 posts I see 3 buttons which is strange

radar, line 12: @bodrovis can you show me a screenshot of what this used to look like?

bodrovis, line 12: Here you go. I have 1 unapproved post but still 2 buttons: one from this view and one from the partial with the post and its moderation tools:

radar:
Added to rails3 and rails4. Thanks!

PR #556, opened 2014-04-04
Remove duplicate friendly_id dependency

bricker:
This duplicate dependency has been in the gemspec for months, is it a mistake or is there a reason? It causes an "invalid gemspec" error when installing.


bodrovis:
https://github.com/radar/forem/pull/554

bricker:
Heh, thanks @bodrovis

PR #565, opened 2014-05-15
Redirect posts#show to topics#show

alexcwatt:
This is to fix #557. (Last couple of commits.)

However, I'm having a bit of trouble with git, as you can see. This pull request is listing too many commits and I can't figure out how to fix that. Any tips?

PR #576, opened 2014-07-10
Redirect posts#show to topics#show

### LINKED ISSUE #557: Route for particular post
I am working on a custom email notifier and tried using the following to get a URL for a particular post that needed to be approved:

```
<%= link_to "Post on #{post.topic}",
forem.forum_topic_post_url(post.forum, post.topic, post) %>
```

It returns a link, but the link does not work... If `/edit` is appended to the URL, it does work.

Making this work will require defining the `show` action for posts. I propose redirecting to the topic path with the appropriate page parameter.

Does this sound like a good approach? Or maybe I'm missing a feature that forem already has?

Thanks.

alexcwatt:
This is to fix #557. It's the same as what I had in #565 except that to get my branch working again, I had to do a hard reset.

### DIFF
--- a/app/controllers/forem/posts_controller.rb
+++ b/app/controllers/forem/posts_controller.rb
@@ -1,14 +1,20 @@
 module Forem
   class PostsController < Forem::ApplicationController
-    before_filter :authenticate_forem_user
+    before_filter :authenticate_forem_user, except: :show
     before_filter :find_topic
     before_filter :reject_locked_topic!, :only => [:create]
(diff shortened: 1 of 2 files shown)


alexcwatt:
@radar Should be ready for merge. I noticed earlier today that the spec I had included had a mistake and was failing. It is fixed now, though some of the jobs in the Travis build still failed for other reasons.

radar:
Thanks @alexcwatt. Looks good to me :)

PR #579, opened 2014-07-27
Remove non-working example links from bottom of Readme

tomhanlon:
First few links in Readme no longer link to working forum examples


radar:
Thank you @tomhanlon

PR #584, opened 2014-08-14
NilUser inherits from Forem.user_class

alexcwatt:
Topic#set_first_post_user expects Forem.user_class

### DIFF
--- a/app/models/forem/nil_user.rb
+++ b/app/models/forem/nil_user.rb
@@ -1,5 +1,5 @@
 module Forem
-  class NilUser
+  class NilUser < Forem.user_class
     def forem_email
       "nobody@example.com"
     end
--- a/spec/models/topic_spec.rb
+++ b/spec/models/topic_spec.rb
@@ -54,6 +54,13 @@ describe Forem::Topic do
     end
   end
 
+  describe ".set_first_post_user" do
+    it "should handle deleted user" do
+      topic.user_id = nil
+      topic.save
+    end
+  end
+
   describe "helper methods" do
     describe "#subscribe_user" do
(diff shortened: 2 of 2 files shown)


radar:
@alexcwatt Please provide steps to reproduce a bug that you were seeing.

alexcwatt:
@radar Sure, I should have done that.
1. Create a new topic.
2. Destroy the user who created the topic.
3. Attempt to load the topic page. You should see an exception because `Topic#set_first_post_user` expects `Forem.user_class` (e.g., `User` in many applications), not `Forem::NilUser`.

radar:
Thanks @alexcwatt. Would you be able to write a regression spec for this as well? I think it should be easy enough to do.

coalwater:
This does fix #609, I would try to write the tests but I'm unable to run rspec due to missing folders that were deleted in a previous commit.

radar:
@coalwater Which missing folders are you talking about?

Travis runs:

```
bundle install
bundle exec rake forem:dummy_app --trace
bundle exec rspec spec
```

And it seems to be able to run the tests OK. What problems are you having?

coalwater:
@radar, Ok i figured out what I was doing wrong, i have 2 folders, 1 for the gem itself and one for a dummy project that uses the gem, I was running the spec on the gem it dummy project, I tested it in the gem folder and I got it working, sorry for the confusion

alexcwatt:
@radar I have (at last) added a regression spec; please let me know if it's what you're looking for. I was not able to reproduce it in exactly the same way, but I think the test captures the issue at a more fundamental level anyway.

radar:
Thanks @alexcwatt, I have verified today that this is an issue. I'll merge your PR now.

PR #595, opened 2014-09-11
Update index.html.erb

montanalow:
Fix the display of the user to use forem_name (optionally with link) instead of a raw object.to_s dump.


radar:
Thank you @montanalow!

PR #599, opened 2014-09-17
Sort forums and categories by order

jengel3:
Unfortunately, there really isn't a way to make sure that your forums and categories are sorted in the order that you want. To make this possible, I've added a order field to categories and forums, which is the order that they will be sorted in.

I needed this, so if it's useless here, then feel free to close it, or let me know of any issues :smile:

### DIFF
--- a/app/controllers/forem/admin/categories_controller.rb
+++ b/app/controllers/forem/admin/categories_controller.rb
@@ -35,7 +35,7 @@ module Forem
       private
 
       def category_params
-        params.require(:category).permit(:name)
+        params.require(:category).permit(:name, :order)
       end
 
       def find_category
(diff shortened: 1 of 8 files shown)


radar:
Hi @Jake0oo0, apologies for the late merge. This looks pretty useful. I'll add it!

radar:
Actually, please give these parameters a different name than "order", as that's a reserved word in most SQL engines.

jengel3:
Alright, will do so when I get a chance!

radar:
Thank you :)

ivgames:
Hah! i made it too)) where were your fixes before))

jengel3:
Any idea what to name it instead @radar? I'm having trouble finding a word to replace it...

yan-hoose:
How about "sort_order"?

radar:
acts_as_list uses "position", so I would just go with that.

jengel3:
#611

PR #600, opened 2014-09-18
Rails 4.1.6 support.

Loremaster:
I added support for Rails 4.1.6.


radar:
Thank you @Loremaster :)

Issue #607, opened 2014-10-03
Error in moderation page


ivgames:
When i try to go to moderation tool, link like http://localhost:3000/forum/testovyy-forum-1/moderation
it thows me this:

undefined method `post_time_tag' for #<#Class:0xb2fef88c:0xb2feee50>

a href='#post-<%= post.id %>'><%= post_time_tag(post) %

TO FIX THIS, need add this: helper 'forem/topics' to moderation_controller.rb after helper 'forem/posts'

radar:
@RostDetal Please submit a patch to fix this issue.

ivgames:
Ок, will do it.

PR #608, opened 2014-10-06
Moderation tools fix

ivgames:
As promised, fix for moderaton tools


radar:
Related to #607.

radar:
@RostDetal Could you please also write a regression test for this issue?

radar:
Ping @RostDetal.

PR #611, opened 2014-10-15
Sort forums and categories by order (Changed to position)

jengel3:
Fixed #599, let me know if you want it rebased.


db/migrate/20140917034000_add_order_to_categories.rb
```diff
@@ -0,0 +1,5 @@
+class AddPositionToCategories < ActiveRecord::Migration
```
radar: This migration file is called `add_order_to_categories`, but the class defined within is `AddPositionToCategories`. You should rename the migration file to the correct name, because this is currently breaking the tests on Travis.

db/migrate/20140917201619_add_order_to_forums.rb
```diff
@@ -0,0 +1,5 @@
+class AddPositionToForums < ActiveRecord::Migration
```
radar: You will need to rename this migration also.

radar:
Thanks for the work @Jake0oo0. Just those two small changes and I think we're good to go (if Travis says so!)

jengel3:
Damn,  I saw the test failed but couldn't figure out why. Will fix in a few.

jengel3:
Should be good @radar, waiting for Travis to finish.

jengel3:
Fixed a broken test...

jengel3:
Woot looks good @radar, failing tests are unrelated to the PR.

radar:
Thanks @Jake0oo0!

bgadoury:
Please excuse me while I flaunt my Gemfile/bundler ignorance, but I ran into a "NoMethodError - undefined method `position'..." error when I re-installed my bundle and pulled this newer commit.

Eventually, I tracked it down to new migrations that I didn't know about. Running them fixed the errors, of course. Should I have run rake forem:install:migrations "just in case" after I re-built my bundle? Or, in other words, is that the "right way" for me to update my bundle? Thanks!

radar:
Yup, whenever you update your bundle you'll need to copy across any new forem migrations and run them.

PR #612, opened 2014-10-15
added translation to the word New

mib32:
Hi, so there wasn't any translation for the word 'New', maybe we should do it?

### DIFF
--- a/app/helpers/forem/topics_helper.rb
+++ b/app/helpers/forem/topics_helper.rb
@@ -13,7 +13,7 @@ module Forem
 
         if forum_view
           if topic_view.nil? && topic.created_at > forum_view.past_viewed_at
-            content_tag :super, "New"
+            content_tag :super, t('forem.general.newly_created')
           end
         end
       end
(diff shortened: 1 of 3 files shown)


app/helpers/forem/topics_helper.rb
```diff
@@ -13,7 +13,7 @@ def new_since_last_view_text(topic)
 
         if forum_view
           if topic_view.nil? && topic.created_at > forum_view.past_viewed_at
-            content_tag :super, "New"
+            content_tag :super, t('forem.general.newly_created')
```
radar, line 16: Since this is for a topic, perhaps we could name the key `forem.topic.newly_created`. WDYT?

jengel3, line 16: That would probably be best, considering there could be a different translation for a feminine/masculine word, depending on the language.

radar, line 16: Great, could you please update this PR for that?

radar:
Ping @Jake0oo0.

PR #616, opened 2014-11-02
Missing branch specification in the documentation

joom:
Forem doesn't have a master branch, so the branch name should be specified.


radar:
Thanks @joom!

PR #618, opened 2014-11-03
[b] Fix migrations

hieuk09:
- ~~No column name when run add_forem_view_fields~~
- No column title when rollback to rename_title_to_name_on_forem_forums then migrate again


radar:
Hi @hieuk09, how do I reproduce this problem on my machine?

hieuk09:
Hi @radar 
- It seems that the first one is already fixed in rails4 branch. It hasn't been fixed on rails 3 yet. I'll remove that change.
- You can reproduce (2) by following these steps:
  - Run `rake forem:dummy_app`
  - Go to spec/dummy
  - run `rake db:rollback STEP=8 RAILS_ENV=test`
  - run `rake db:migrate`

radar:
Sorry about the delay. I've been busy writing. 

This patch does work for me. Thanks very much :+1:

PR #619, opened 2014-11-04
Fix a few sorting issues

yan-hoose:
Currently forum catgories are sorted "position DESC" in ForumsController#index, which is the opposite from the default forum sorting. Also, forum position can't be set on the forum form in the admin interface, since :position is not permitted.

This commit does this:
- default Category sorting to "position ASC", same as with the Forum model
- allow forums "position" to be mass-assigned

### DIFF
--- a/app/models/forem/category.rb
+++ b/app/models/forem/category.rb
@@ -9,6 +9,8 @@ module Forem
     validates :name, :presence => true
     validates :position, numericality: { only_integer: true }
 
+    default_scope { order(:position) }
+
     def to_s
       name
     end
--- a/app/models/forem/category.rb
+++ b/app/models/forem/category.rb
@@ -9,7 +9,7 @@ module Forem
     validates :name, :presence => true
     validates :position, numericality: { only_integer: true }
 
-    default_scope { order(:position) }
+    scope :by_position, -> { order(:position) }
 
     def to_s
       name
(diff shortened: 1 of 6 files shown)


radar:
Hi @yan-hoose, thanks for the patch.

Could you please add a regression test for the `Admin::ForumsController`?

yan-hoose:
Hi. Test added!

app/models/forem/category.rb
```diff
@@ -9,6 +9,8 @@ class Category < ActiveRecord::Base
     validates :name, :presence => true
     validates :position, numericality: { only_integer: true }
 
+    default_scope { order(:position) }
```
radar: I'm always wary of `default_scope` usage... could we make this a scope instead and say `Forem::Category.by_position` instead? Sorry to be a pain.

bodrovis: I too think that making this a scope would be better

yan-hoose: Sure, why not, I'll make the changes. Should I start from scratch and create a new pull request for that?

radar: You can just do a force push to this branch and the PR will be updated automatically.

yan-hoose:
Replaced the default scope with a by_position scope.

radar:
Thanks @yan-hoose! This has been merged.

PR #620, opened 2014-11-14
polish translation equal to english

h3xed:
Supplemented Polish translation.


radar:
Thanks very much @h3xed :) :tada: :beers:

PR #623, opened 2014-11-23
Pass option to forum_pages_widget helper

sungwoncho:
The view files in forem uses `forem_pages_widget` helper to render pagination, but there is no way to pass in additional arguments. I thought adding arguments might be useful.

For instance, such need arises when using will_paginate-bootstrap to style pagination rendered by will_paginate. People need to pass `renderer: BootstrapPagination::Rails` to will_paginate helper, but currently there is no easy way to do so.

### DIFF
--- a/app/helpers/forem/application_helper.rb
+++ b/app/helpers/forem/application_helper.rb
@@ -19,7 +19,7 @@ module Forem
     def forem_pages_widget(collection, options={})
       if collection.num_pages > 1
         content_tag :div, :class => 'pages' do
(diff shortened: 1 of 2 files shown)


radar:
Seems good to me :) Can you please write a test for this so that we can ensure this behaviour sticks around?

sungwoncho:
Hi @radar. I finally got around writing specs for this. Let me know what you think.

sungwoncho:
ping @radar

radar:
Thanks @sungwoncho. LGTM.

PR #627, opened 2014-12-16
Added Estonian translations and alphabetized the language list

yan-hoose:


radar:
Thank you very much @yan-hoose :)

PR #628, opened 2014-12-22
For tests, use a newer version of Devise that supports Ruby on Rails 4.2.

derikson:
Without this change, `bundle exec rspec spec` on Ruby on Rails 4.2 results in an error with no tests being run.

With this change, the tests run, albeit with a few failing tests and a load of deprecation messages.


radar:
Thanks @derikson!

PR #632, opened 2015-01-16
validate length of topics subject

jlerpscher:
Avoid 500 error when topics subject is to long for string type


radar:
Thanks @jlerpscher!

PR #633, opened 2015-01-16
add missing translations in french

jlerpscher:


radar:
Thanks for these too @jlerpscher!

PR #634, opened 2015-01-22
Add dependent => destroy to forum/topic views and topic subscriptions

yan-hoose:
To not leave behind orphaned subscription and view rows when you delete a topic or a forum or to not get a database exception when you are using foreign keys in your Forem tables, I added dependent => destroy options to topic subscriptions and viewable views.


radar:
Looks good to me. Thanks @yan-hoose!

Issue #637, opened 2015-02-11
dependency on simple_form ~> 3.0.1 - 3.1.0 needed


tomg65:
Hi,

I tried to integrate forem in an existing rails apllication (rails 4.1.9 ruby 2.1.3)
There is already a cms engine (https://github.com/AlchemyCMS/alchemy_cms) integrated that depends on simple_form 3.1.0.

So I can't install forem in this application.

Is it possible to upgrade the dependency in forem?

Thanks
Thomas

radar:
I don't understand. Why can't you install Forem into this application?

radar:
If all it is is the `simple_form` dependency in the gemspec, why not submit a patch to change it to something more like:

``` ruby
s.add_dependency 'simple_form', '~> 3.0'
```

I think that would seem pretty reasonable.

tomg65:
bundler stops while showing a conflict message in combination with the other gem, which is dependent on simple_form 3.1
OK - changed the dependency in your gemspec. and made a pull request #639 
With this adaption bundle install works.

radar:
Done already I believe.

PR #639, opened 2015-02-12
update of simpleform dependency in gemspec

tomg65:
bundler stops while showing a conflict message in combination with other gem, which is dependent on simple_form 3.1 ( issue #637 )


radar:
Could you please make this a `~> 3.0` dependency instead?

tomg65:
OK - made `~> 3.0`
:)

radar:
Thanks!

PR #648, opened 2015-03-22
Add Korean translation

sungwoncho:


radar:
Thank you!

yan-hoose:
Hey, I just noticed that there is "en" as the locale as the first line in the ko.yml file.

PR #649, opened 2015-03-24
Fixes bug where Korean text overrides English text

colinrymer:
See https://github.com/radar/forem/pull/648#issuecomment-85225320 for additional context.


radar:
Thanks. Sorry for that :(

sungwoncho:
Oops.. Good catch.

Issue #640, opened 2015-02-15
Rails 4.2 support (waiting on rails/rails#19042)


yan-hoose:
# [Ed (Radar): We are waiting for this bug to be fixed in rails: rails/rails#19042]

Hi,

I'd like to see Forem updated to support Rails 4.2. I took a look at it and at first glance there does not seem to be very much needed to do. There are a ton of deprecation warnings when running specs but most of them are repetitive. However, there are a few strange issues.

BTW, currently Travis CI is running Forem tests in Rails 4.2 mode, so you can see the warnings easily yourself. In Rails 4.2 mode because in the gemspec there is currently
`s.add_dependency 'rails', '~> 4.0', '<= 4.2'`
instead of
`s.add_dependency 'rails', '~> 4.0', '< 4.2'`.

Anyway, when running tests, there is one failing test and three main sources of warnings:
- deprecated `use_route` in `forem/spec/support/controller_hacks.rb:27`
- using `#deliver` instead of `#deliver_now` or `#deliver_later` in `forem/app/models/forem/subscription.rb:12`
- `#column_for_attribute` change warnings from Simple Form: `forem/app/views/forem/admin/forums/_form.html.erb:8`. This warning will be resolved if the gem dependency is changed from '~> 3.0.1' to '~> 3.0'.

I'm not sure if in addition to fixing the `use_route` deprecation warning, RSpec gem dependency should be raised from '~> 2.14.0' to '~> 3.1.0' (since 3.1 is the one with official Rails 4.2 support). Somebody with more experience in RSpec might know the answer to this.

There is also a question about how to support both #deliver and #deliver_later depending on the Rails version.

About the failing test:
`rspec ./spec/models/topic_spec.rb:87 # Forem::Topic helper methods #register_view_by increments the overall topic view count`
It seems that this code in forem/models/forem/concerns/viewable.rb

``` ruby
view = views.find_or_create_by(user_id: user.id)
view.increment!('count')
increment!(:views_count)
```

is incrementing topic.views_count by 2 instead of 1 - first increment in `views.find_or_create_by` (I'm not sure why) and the second one in `increment!(:views_count)`. 

Has anybody else taken a look at 4.2 support and knows the answers/solutions to these issues already? Unfortunately I don't have too much time to spend on this.

radar:
Hi yan-hoose,

If you'd like to see it updated then the best way to get that to happen would be to submit a PR which did the updating.

Could you please work on that? I will review it when you're done.

Thanks

[...]

PR #659, opened 2015-05-14
able to use with rails 4.2.1 too

mpakus:


yan-hoose:
Since there is a bug in Rails 4.2 that causes Forem to double-increment views, I don't recommend to use Forem with Rails 4.2 until that bug is fixed (https://github.com/radar/forem/issues/640).

If you are fine with that, go ahead! :smile:

radar:
I am not fine with having Forem support a version of Rails with a known regression. Therefore I won't be accepting this PR until Rails 4.2.2 is released.

senych-vitalii:
hey @radar Rails 4.2.2 is already released, can you accept PR?
Currently I cannot use Forem without this fix because I use 4.2.1 version of Rails...

radar:
Does 4.2.2 contain the specified patch or is it just security releases?

senych-vitalii:
@radar it seems not... 
Sorry, I missed that counter is working in 4.2 version and is not working in 4.2.1.

yan-hoose:
Actually counter caches are broken in all of Rails 4.2 versions. Forem Rails dependency should be

``` ruby
s.add_dependency 'rails', '~> 4.0', '< 4.2'
```

instead of

``` ruby
s.add_dependency 'rails', '~> 4.0', '<= 4.2'
```

radar:
I'm fine with leaving the dependency as it is now. If people complain about the counter cache issue, we can point them to this issue.

edmondchui:
The counter issue is fixed in Rails 4.2.4. Could you please update the dependency to Rails 4.2.4, 4.2.5, and above? Thanks.

sintro:
> The counter issue is fixed in Rails 4.2.4. Could you please update the dependency to Rails 4.2.4, 4.2.5, and above? Thanks.

Is not it already updated?

PR #660, opened 2015-05-14
Reduce the Number of Migration Files

TekuConcept:
I noticed when adding the forem engine to our website that there were a total of 32 migration files generated. I merged them into 9 individual table files and 1 user file.


TekuConcept:
My bad, I stumbled upon the 'CONTRIBUTING.md' file several hours after creating this pull request. I am currently in the process of fixing those errors presented in the 'rspec spec' tests.

yan-hoose:
If you are talking about the Travis CI test failures then all of them are currently caused by Forem specifying

``` ruby
s.add_dependency 'rails', '~> 4.0', '<= 4.2'
```

in the gemspec even though there is no Rails 4.2 support right now. This means that the tests are run in the context of a Rails 4.2 app and results in a lot of errors. Rails 4.2 support and all the errors have been fixed but not yet pushed to Forem since there is a bug in Rails 4.2 that prevents the push. More details here: https://github.com/radar/forem/issues/640

TekuConcept:
Good to know; I will push my updates based on the contribution standards.

Issue #664, opened 2015-06-18
restful API?


eostrom:
I need to add a forum to an existing Rails app with an Angular front end. An option I'm considering is adopting Forem (which I've used before), taking advantage of its data model and possibly URL structure, and adding a JSON API. Does this sound like a good or bad idea to anyone?

And have you given any thought to an approach? It looks like the best options are either to integrate with the existing routes/controllers and add JSON views (maybe using RABL) or to build out a separate HTTP/URL structure (using something like Grape). I'm inclined to use RABL because that's what we're doing in the main engine of my application, but if there's a strong preference from the team, I'd rather do something that's likely to get integrated into Forem.

radar:
For about a year now I've thought that Forem would be much better if it was just a JSON API without a frontend at all. A lot of people just customise the frontend or ditch it entirely. Most people don't use what we supply. I had a similar set of thoughts when I worked at Spree too.

I haven't had the time to invest in doing the work because I've been too busy with my full-time job and writing books.

I think it sounds like a step in the right direction. We should not use Rabl for this as I've had issues with it ; in my experience with building a Rabl-backed API for Spree. Grape is not that great either. I'm fine with it being straight jbuilder views in Rails. 

I think it should comply with [the JSON API specification](http://jsonapi.org/). It should be versioned and access should be granted with user-based API keys (i.e. each user can have an API key generated to access Forem). 

Please do it :)

eostrom:
What are your thoughts on view counts? In an API-based app, there isn't necessarily a one-to-one correspondence between a browser requesting data for a topic and a user viewing that topic. Should we bother having an explicit "views" API where a client can let the server know what the user's been looking at?

radar:
I think if a topic has been retrieved directly (i.e a request like /forums/blah/topics/foo) then that should count as a view. Having a separate endpoint for this doesn't make much sense imo.

What's you're reasoning for having a separate endpoint?

eostrom:
I forget. I mean, I can imagine building an app in various ways that request a forum endpoint with included topic data, or that request topics and don't necessarily render them. But I'm not currently doing that, so maybe I was just overthinking it.

vfonic:
Could it be an optional url parameter for such cases when you want to preload some forums/topics?
For example: /forums/blah/topics/foo?bump_view_counter=false

PR #665, opened 2015-07-02
API proof of concept: forums#show

eostrom:
THIS IS NOT A REAL PULL REQUEST AND SHOULD NOT BE MERGED. It's just the easiest way for me to illustrate the approach I'm using so far and ask for feedback. This follows discussion in radar/forem#664.
- route is /api/forums/:id (may be slug)
- versioning is done via Accept header (roughly on the [Pat Allan model](http://freelancing-gods.com/2011/09/24/versioning-your-ap-is.html))
- format is [JSON API](http://jsonapi.org/)
- views are built with [jbuilder](https://github.com/rails/jbuilder/)
- authentication is done via session cookies (or whatever the host app is using)

@radar mentioned API keys for authentication, but I don't need them for my app, so I'm keeping it simple.

I'm writing request specs, and not feature or controller specs. It just seems like the natural level for testing this functionality.

I haven't thought about how to represent errors (or read about how JSON API does it). Nor have I started figuring out associations.

I created a separate controller hierarchy, rather than try to cram versioned API functionality into the regular controllers. I am not 100% sure about this.

I am also not sure it wouldn't be better to use [JSONAPI::Resources](https://github.com/cerebris/jsonapi-resources) instead of creating the views manually with jbuilder.

### DIFF
--- a/app/controllers/forem/api/v1/forums_controller.rb
+++ b/app/controllers/forem/api/v1/forums_controller.rb
@@ -3,9 +3,7 @@ require_dependency "forem/application_controller"
 module Forem
   module Api
     module V1
-      class ForumsController < ApplicationController
-        load_and_authorize_resource :class => 'Forem::Forum', :only => :show
-
+      class ForumsController < Forem::ForumsController
         rescue_from CanCan::AccessDenied, with: :forbidden
 
         before_action { request.format = :json }
--- a/app/controllers/forem/api/v1/forums_controller.rb
+++ b/app/controllers/forem/api/v1/forums_controller.rb
@@ -4,15 +4,7 @@ module Forem
   module Api
     module V1
       class ForumsController < Forem::ForumsController
-        rescue_from CanCan::AccessDenied, with: :forbidden
-
-        before_action { request.format = :json }
-
-        private
-
-        def forbidden
-          render nothing: true, status: :forbidden
-        end
+        include JsonApiController
       end
     end
   end
--- a/app/controllers/forem/api/v1/forums_controller.rb
+++ b/app/controllers/forem/api/v1/forums_controller.rb
@@ -6,7 +6,15 @@ module Forem
       class ForumsController < ApplicationController
         load_and_authorize_resource :class => 'Forem::Forum', :only => :show
 
+        rescue_from CanCan::AccessDenied, with: :forbidden
+
         before_action { request.format = :json }
+
+        private
+
+        def forbidden
+          render nothing: true, status: :forbidden
+        end
       end
     end
   end
--- a/app/controllers/forem/api/v1/forums_controller.rb
+++ b/app/controllers/forem/api/v1/forums_controller.rb
@@ -1,0 +1,13 @@
+require_dependency "forem/application_controller"
+
+module Forem
+  module Api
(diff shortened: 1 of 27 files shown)


app/controllers/forem/api/v1/forums_controller.rb
```diff
@@ -0,0 +1,13 @@
+require_dependency "forem/application_controller"
+
+module Forem
+  module Api
+    module V1
+      class ForumsController < ApplicationController
+        load_and_authorize_resource :class => 'Forem::Forum', :only => :show
+
+        before_action { request.format = :json }
```
radar: I thought that this kind of request format setting could be done at the routes level but I can't find a solution that works. I tried coming up with my own and googling for one. Nothing that I tried works.

Instead, putting `respond_to :json` at the top of the controller works, but this requires the use of the `responders` gem (which is alright imo). I think this is a better method than forcing the request format this way.

forem.gemspec
```diff
@@ -31,4 +31,5 @@ Gem::Specification.new do |s|
   s.add_dependency 'select2-rails', '~> 3.5.4'
   s.add_dependency 'friendly_id', '~> 5.0.0'
   s.add_dependency 'cancancan', '~> 1.7'
+  s.add_dependency 'jbuilder'
```
radar: I would personally lock the jbuilder version here, just in case of a major version bump that changes things in a breaking way.

eostrom:
Thanks, I'll make those changes next.

I've just pushed four new commits that contain tentative answers to questions I ran into:
- [Can an API user create a Topic without a Post?](https://github.com/eostrom/forem/commit/7e8ad68548d1f8ca1a607b69d71464ff1ad9d081)
- [How do we keep `/api/` routes from colliding with a forum about APIs?](https://github.com/eostrom/forem/commit/c5a78c3d2faf0ab7104b2c4694cc1b5b61259220)
- [How closely should API controllers be integrated with HTML UI controllers?](https://github.com/eostrom/forem/commit/88ca5e727d033b9786b1dbf8bef3f8ca746e42aa)
- [How should we represent errors in API responses?](https://github.com/eostrom/forem/commit/aa614d91e624f8761f734dbc3fe5aad47f7fce4e)

I'm pressing ahead with these tentative answers, but let me know if you have thoughts on them.

app/controllers/forem/api/v1/forums_controller.rb
eostrom: `respond_to` doesn't seem to address the problem I was trying to solve. The problem is that I'm using a custom MIME type (`application/vnd.forem+json`), as recommended in any number of blog posts, but because of that, Rails looks for a view named `show.application/vnd.forem+json.jbuilder`. I just wanted to call it `show.json.jbuilder`.

I tried `Mime::Type.register 'application/vnd.forem+json', :json`, which works, but I get a warning about initializing `Mime::JSON` twice. There doesn't seem to be a way to say "Just treat this MIME type as JSON, too." (`register_alias` doesn't do it.)

It looks like a solution is to register the MIME type as `:forem` and then rename the views to `show.forem.jbuilder`. But it seems like there should be a way to use `.json`.

PR #668, opened 2015-07-12
Rails 4.2 support

yan-hoose:
I thought that I'd go ahead and post the PR even though the needed Rails fix is not present yet. This way you can already start reviewing the code (who knows when the Rails fix will happen). This is what got done:
- upgrade RSpec from 2.14 to 3.2 including upgrading from should to expect syntax
- use deliver_later when the mail object responds to it (starting from Rails 4.2)
- replace deprecated STDOUT capturing with alternatives
- add a default simple_form config to the dummy app to get rid of warnings
- add explicit :null => true to timestamps migrations (the current default) since it will be :null => false by default in future Rails versions

Tests pass nicely in the context of Rails 4.1. However in the context of Rails 4.2 there is one test failure caused by the known Rails counter_cache bug. Otherwise tests run without warnings or deprecation messages.


app/models/forem/subscription.rb
```diff
@@ -9,7 +9,12 @@ def send_notification(post_id)
       # If a user cannot be found, then no-op
       # This will happen if the user record has been deleted.
       if subscriber.present?
-        SubscriptionMailer.topic_reply(post_id, subscriber.id).deliver
+        mail = SubscriptionMailer.topic_reply(post_id, subscriber.id)
```
radar, line 12: I'm considering dropping support for earlier versions of Rails so that we won't have this method at all. It will always be `deliver_later`, and then `ActiveJob` can handle it.

spec/configuration_spec.rb
```diff
@@ -4,20 +4,20 @@
   describe ".default_gravatar" do
     it "can be set and retrieved" do
       Forem.default_gravatar = "foo"
-      Forem.default_gravatar.should eq("foo")
+      expect(Forem.default_gravatar).to eq("foo")
```
radar, line 7: Thanks for bringing this up to scratch on RSpec 3 :)

spec/lib/generators/forem/dummy/templates/config/initializers/simple_form.rb
```diff
@@ -0,0 +1,166 @@
+# Use this setup block to configure all options available in SimpleForm.
+SimpleForm.setup do |config|
```
radar, line 3: I don't think we should include this by default. SimpleForm already defaults to its own default configuration if one isn't specified. Why do we need to specify it ourselves?

yan-hoose, line 3: I added it because it was outputting warnings about missing config (and the fact that simple_form will use its defaults) when running the specs.

radar, line 3: My concern would be that if `simple_form` changed their defaults, then these "defaults" that we have here would be out of date, and therefore incorrect.

yan-hoose, line 3: I did some testing and it seems when we specify an absolute minimum config block like this:

``` ruby
SimpleForm.setup do |config|
end
```

... then simple_form does not complain about missing config. Sounds good, should I make the changes?

radar, line 3: Does that still use the default conf?

yan-hoose, line 3: Yes, this empty block gives me the same config as without any config block at all.

radar, line 3: Wonderful!

yan-hoose:
Well, everything is green!

radar:
:shipit:

PR #670, opened 2015-07-16
Provide a workaround for failing spec

vfonic:
I'm not sure this is the best way to go around this, but it does solve the issue without the need to change any functionality (e.g. changing column name).

The solution relies on the fact that counter_cache fails silently when there's no column with given name.


radar:
I think I will rather wait for Rails 4.2.4 to be released. Thank you for investigating however!

PR #673, opened 2015-08-07
upgrade rails version and fix bug

wuwx:
Rails 4.2 => 4.2.3
Forem::Topic helper methods #register_view_by increments the overall topic view count

### DIFF
--- a/app/models/forem/subscription.rb
+++ b/app/models/forem/subscription.rb
@@ -9,7 +9,7 @@ module Forem
       # If a user cannot be found, then no-op
       # This will happen if the user record has been deleted.
       if subscriber.present?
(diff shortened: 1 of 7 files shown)


radar:
Please see #640. We're waiting on 4.2.4 to be released.

app/models/forem/subscription.rb
```diff
@@ -9,7 +9,7 @@ def send_notification(post_id)
       # If a user cannot be found, then no-op
       # This will happen if the user record has been deleted.
       if subscriber.present?
-        SubscriptionMailer.topic_reply(post_id, subscriber.id).deliver
+        SubscriptionMailer.topic_reply(post_id, subscriber.id).deliver_now
```
radar, line 12: And what if people want to use `deliver_later` instead?

Issue #675, opened 2015-08-20
Rails 4.1.1 development class reloading issue


sr3d:
We are trying to integrate Forem into a Rails4.1.1 app but so far we're running into issues with the reloading of the Forem classes in development.  For example,

1st request is OK -- we can see the forem page.
2nd request we'd get an error like 

NameError in Forem::Categories#show
Showing /Users/alex/.rvm/gems/ruby-2.0.0-p247/bundler/gems/forem-c793fb195ad4/app/views/forem/forums/_forum.html.erb where line #7 raised:

uninitialized constant Forem::Forum::Post

Disabling the cache reloading in config/environments/development.rb (config.cache_classes = true) fixes the issue since now the classes don't get reloaded.

We're using Forem off the rails4 branch already:

gem 'forem', :github => "radar/forem", :branch => "rails4"

neocanable:
+1
we got same things

radar:
Please show me your `Gemfile`.

cjbottaro:
+1
Same issue here.  For me, every time it's `uninitialized constant Forem::StateWorkflow`.  I have to restart my server almost every time I make a change.  Development is extremely painful now.

Gemfile:  https://gist.github.com/cjbottaro/4373517f4a8efcb6e3e1

radar:
Any particular reason why you're stuck on 4.1.1 and can't upgrade to 4.2.4?

cjbottaro:
I'm not sure what you mean.  The Gemfile that I posted shows 4.2.4.  In fact, I had to upgrade to 4.2.4 (from 4.2.3) just to get it to work (due to dependencies specified in the gemspec).

radar:
You're commenting on an issue that says it has a problem with 4.1.1. You said you have the same issue.

I didn't see that your Gemfile contained 4.2.4.

cjbottaro:
Whoops, sorry about that.  I meant to say I'm also experiencing this with Rails 4.2.4.

radar:
Could someone please share with me a basic application which reproduces this issue? I'm having a hard time reproducing it myself.

radar:
This is fixed now by #679.

PR #679, opened 2015-09-01
Fixes class reloading issue

cjbottaro:
This fixes the issue for me.  Sorry, couldn't run specs because I don't want to install MySQL, but it seems like a pretty innocuous change.

Pretty sure bug was caused by something related to this:
http://guides.rubyonrails.org/autoloading_and_reloading_constants.html#autoloading-and-require

Something about autoloading classes (in `app/`) requiring non-autoloading classes (in `lib/`).  ¯_(ツ)_/¯

### DIFF
--- a/app/models/forem/post.rb
+++ b/app/models/forem/post.rb
@@ -1,5 +1,3 @@
-require 'forem/state_workflow'
-
 module Forem
   class Post < ActiveRecord::Base
     include Workflow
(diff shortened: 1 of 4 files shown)


cjbottaro:
Fixes issue #675 for me.

radar:
Haha this is great! I was just warning another developer about doing this kind of thing in a model and then it's pointed out that I've done it myself!!

The require call is only ran once, and so on the second request when it reloads the model it won't run it again. The solution is to do as you've done, or to require them with require_dependency which uses load or require depending on a config setting I forget the name of right now.

Thanks very much for the patch! Travis has run the tests and according to it the tests are fine.

I'll merge now!

PR #683, opened 2015-09-21
Rails3

archievi:


radar:
No description and failing build and it doesn't merge cleanly. Assuming PR opened in error.

PR #686, opened 2015-11-01
Set mysql version in Gemfile to work with latest rails

zakak:
This should fix the failing mysql Travis builds until the latest mysql2 + rails get along. http://stackoverflow.com/a/32466950


zakak:
mysql jobs passed, but the [tests are failing with a particular seed](https://travis-ci.org/radar/forem/jobs/88627403). Reproducible with any DB so I created a separate issue #687.

